PR #3, opened 2020-05-05
Added a function to get a map of a kazoo documents setter functions.

bradfordben:
Adding this for use by kazoo tasks. This will allow tasks to dynamically get a list of attributes it can set on a user from kazoo_documents.

I will pr the users kazoo tasks module in its repo.


kazoo_documents/src/kzd_module_utils.erl
```diff
@@ -30,6 +31,26 @@
...
        ).
 
+%%------------------------------------------------------------------------------
+%% @doc Get a mapping of all setter functions (start with set_) with arity 2 for
+%% a document.
+%% Returns map where the key (Function name without set_) is a binary and the
+%% value (Function name) is an atom.
+%% @end
+%%------------------------------------------------------------------------------
+-spec get_setters(atom()) -> map().
```
jamesaimonetti: Can you spec the `map()`? Maybe `#{kz_term:ne_binary() => atom()}` or whatever it should be?

jamesaimonetti (changes requested):

kazoo_documents/src/kzd_module_utils.erl
bradfordben: 👍

jamesaimonetti:
Merged

bradfordben:
@jamesaimonetti Thanks

PR #9, opened 2020-05-19
Modify kapi_conference module to use kapi_definition module

harenson:
Depends on https://github.com/2600hz/kazoo-ecallmgr/pull/5 PR to be merged first and then rebase master and push changes here so it can pass all CI checks.


kazoo_amqp/src/api/kapi_conference.erl
```diff
@@ -654,984 +1440,908 @@ tones(Prop) when is_list(Prop) ->
...
+-spec conference_error_v(kz_term:api_terms()) -> boolean().
+conference_error_v(Req) ->
+    kapi_definition:validate(Req, conference_error_definition()).
+
+-spec publish_conference_error(kz_term:ne_binary(), kz_term:api_terms()) -> 'ok'.
+publish_conference_error(Queue, JObj) ->
+    publish_conference_error(Queue, JObj, ?DEFAULT_CONTENT_TYPE).
+
+-spec publish_conference_error(kz_term:ne_binary(), kz_term:api_terms(), kz_term:ne_binary()) -> 'ok'.
+publish_conference_error(Queue, Req, ContentType) ->
```
lazedo: why the rename ?

lazedo (changes requested):

kazoo_amqp/src/api/kapi_conference.erl
harenson: To match the convention with the other functions, so the event name is `conference_error` so the publish function "should" be named `publish_conference_error` that way you know it is for the `conference_error` event.

lazedo: i don't see that as a good reason for potentially break other people's code. iirc, `conference_error` was named because it couldn't/shouldn't be `error`. it really feels strange having `kapi_conference:publish_conference_error`  because `conference` is already in the `kapi` name

lazedo (approved):

PR #13, opened 2020-05-25
Added the ability to trim a vm length to a defined length before sendingit the audio to the defined asr provider

bradfordben:
Few issues came to light with this addition.
The recording_seconds was actually recording_milliseconds and it was also not being set correctly so i fixed both in here as they are needed.

Requires https://github.com/2600hz/kazoo-crossbar/pull/6

Also submitted to 4.3 at https://github.com/2600hz/kazoo/pull/6546


bradfordben:
Requires https://github.com/2600hz/kazoo-crossbar/pull/6

bradfordben:
Also submitted to 4.3 https://github.com/2600hz/kazoo/pull/6546

bradfordben:
@mk1s Thanks for approving https://github.com/2600hz/kazoo/pull/6546 

Can you approve this pr and https://github.com/2600hz/kazoo-crossbar/pull/6 also for the same feature / code

PR #18, opened 2020-06-01
PISTON-1078: do not automatically update accounts doc when validated

danielfinke:
- the bug allows the doc to be saved even if there are errors
- it also causes there to be 2 saves per crossbar save


lazedo:
imo, this was a bad move https://github.com/2600hz/kazoo-core/commit/f3e6132a6af81a07c9c45d98c0f4a7080ea51e84 . kazoo_documents should be about accessors (getters & setters). considering reverting that and do it the proper way (separation of concerns)

lazedo:
merged this as its fixing a bug introduce with previous accepted pr

we need to consider rolling back as commented above

lazedo:
@danielfinke thank you for the fix

danielfinke:
@lazedo you're welcome
do you have some thoughts as to where to move the validation logic so that it can be shared by any applications that want to make sure they are following the validation rules? I think that was the intent of moving the logic away from crossbar alone.

PR #20, opened 2020-06-03
do not normalize username

lazedo:
this [commit](https://github.com/2600hz/kazoo/commit/d1c3ebea48d38f02ec9bd66887ea728ae30fae9c#diff-cfc207e85505b2fdceef5ce9617dcc1f) broke some external crossbar module by [not honoring the
that username normalization would depending of context](https://github.com/2600hz/kazoo/blob/9770c5e32e5e8628edac1bc6bb5d373abbbc58fb/applications/crossbar/src/modules/cb_users.erl#L507)

there's no real need for normalization, and we should stop doing it in other places


k-anderson:
@lazedo thats not correct, we do need to normalize the username and account name during login... please revert this...

From Rick: FWIW, I've been logging into my own reseller account with "Rick@getclarity.biz" for YEARS. But, this morning I had to log in as "rick@getclarity.biz". Was there some case normalization we used to do that we stopped doing recently?

lazedo:
@k-anderson well, should we revert the one that caused the issue in the first place ? on  `Rick`, what's the actual username ?

PR #23, opened 2020-06-12
add dev.yml partial indexes

jamesaimonetti:


jamesaimonetti:
Leans on https://github.com/2600hz/kazoo5/pull/12

PR #30, opened 2020-07-01
Modify kapi_im module to use kapi_definition module

harenson:


lazedo:
@harenson same comments as in https://github.com/2600hz/kazoo/pull/6363

harenson:
@lazedo I just replied to your comments on the other PR. Let me know if that helps.

icehess:
@lazedo did Harry's comments on the original PR answered your questions?

About macros, I kind of prefer to not use macros if it is not necessary. I prefer to use macros when we are using same value in multiple places or we want to avoid ambiguity or some other special cases. But other than that it just make code reading harder, especially in this case, you have to constantly jumping to macro definition and to places it was used or even you have to open the header file.

Without macros everything is in the same place nice and clean. And if somehow change something there is no unused macros hanging around anymore. Again we can use them to avoid repeating stuff like some of stuff in `kapi_notifications`.

lazedo:
@icehess didn't have time to test. this `kapi` is `critical` and a change without proper testing is not advised at all.
besides that, this `kapi_definitions` really needs some work around, we can talk offline on this.

jamesaimonetti:
@harenson can you check on what's needed here, what tests @lazedo wants to have in place, etc?

harenson:
@lazedo do you need anything from me in order to get this PR merged?

lazedo:
@harenson sorry, as mentioned before, i didn't test this and not going to merge until it i tested it. not a priority atm.

manushiperera:
Hello @harenson,
This PR has conflicts with the latest master branch.
Is this still relevant?

jamesaimonetti:
If needed, we'll figure out how to reopen on the other side. Closing for now

PR #32, opened 2020-07-02
PISTON-1002: remove unnecessary kapps_call_command:collect_digits/8

danielfinke:
kazoo-callflow PR: https://github.com/2600hz/kazoo-callflow/pull/5

- I previously introduced it in 2600hz/kazoo#6249
- usage check using xref:q has been done

I think we should also go over the behaviour of each of the clauses of collect_digits, because they're mostly the same, but it's only the /5 version that properly has a long after_timeout until the noop is returned. It seems like the other clauses would proceed without waiting for a longer media file - the after_timeout should be the 1 day default until the media file noop comes back


danielfinke:
Error is related to the usage that is being removed in 2600hz/kazoo-callflow#5

PR #34, opened 2020-07-06
Rm kazoo

jamesaimonetti:


kazoo_apps/src/kapps_util.erl
```diff
@@ -718,3 +725,36 @@ epmd_enabled() ->
...
+    {'memory', Memory} = erlang:process_info(self(), 'memory'),
+    Memory.
+
+%%------------------------------------------------------------------------------
+%% @doc Fetch and cache the kazoo version from the VERSION file in kazoo's root folder/
+%% @end
+%%------------------------------------------------------------------------------
+-spec kazoo_version() -> kz_term:ne_binary().
+kazoo_version() ->
+    {_, _, Version} = kz_application:get_app('kazoo'),
```
lazedo: it seems you're removing the `kazoo` app, so this won't work

kazoo_apps/src/kapps_util.erl
```diff
@@ -66,6 +66,13 @@
         ,epmd_disabled/0
         ]).
 
+-export([bin_usage/0
+        ,mem_usage/0
+        ,kazoo_version/0
+        ,node_name/0, node_hostname/0
```
lazedo: there's some apps that don't use kazoo_apps and its not even included in the release. can we move this elsewhere ?

lazedo (commented):
only concern is with `kazoo_version` , `node_name` and `node_hostname` in `kapps_util`.
kazoo_apps starts a lot of things that some deployments don't and therefore are not even integrated in the release. with this change, the `auto dependency`  checks will add `kazoo_apps` to those deployments.

`kazoo_maintenance` also has some great `funs` that should be moved elsewhere so they are available in simple deployments (without `kazoo_apps`)

since you created `kz_application` and moved the `calling_app` routines there, maybe move the `hotload_app` there too ? and the `hotload` to `kz_module` ?
or maybe just copy `kazoo_maintenance` as `kz_maintenance` inside `kazoo_stdlib` ? and change the `kazoo_maintenance` to call the kz_maintenance if we want to keep the name ?

PR #36, opened 2020-07-07
Standardize views and implement an abstraction for fetching views

icehess:
# View changes overview

Here is a summary of changes to the views in Kazoo 5. As part of cleaning up, we removed old and used views, and normalized the common tasks of views (like list by DocId in `crossbar_listing` views) in a single uniform views. Also to indexing performance reason system core views are getting move to a single design document as far as possible.

There would be a mechanism for applications to add their own design/views.

Also to Kazoo 5 is providing an abstraction over the way we retrieve the views data. The views accessors has been added as a standard way to access the data. The idea is that this accessors are providing the high level concept of entities and the database drivers are implementing the way to retrieving the data from the database.

```
[kzd modules] --> [kzv modules] --> [kazoo data and storage system] --> [database/storage driver]
```

## Cursor

The new `kz_view` module is an interface to using the view accessors. It returns a cursor when you use `kz_view:find_batch` which then you can use it to do various things with functions from `kz_view_cursor` to fetch data in unlimited, limited or requested batch size, do foreach and mapfold and get the whole view count (or iterate the result and return the count) on the cursor. Other stuff like `reduce` or `aggregation` can be implemented easily in the future.

`kz_view_cursor` will enforce the page limit (if any) and adjust how much tows to fetch. the driver `kz_couch_view` then tries to satisfy that batch limit. This is the same as the old `crossbar_view` in a more clean and formal implementation.

`kz_view` or `kz_view_cursor` can also work with other views, just set `{no_design_module, false}` as option. You have to build and set `startkey`, `endkey` and `databases` options yourself.

Crossbar PR: https://github.com/2600hz/kazoo-crossbar/pull/32

## View Changes

Most of he views are move the `crossbar_listings` design document. This is all in account db right now, but if it is successful we would considering doing the same for all other databases.

Also there were some unused views which has been remove.

## Test

There is test in kazoo-properly (https://github.com/2600hz/kazoo-properly/pull/12) to compare the result of the old view and new view.

## Required PRs

* [merged] fix dialyzer warnings for `kazoo_couch_view`: https://github.com/2600hz/kazoo5/pull/19
* [merged] really fix dialyzer warning for `kz_couch_view`: https://github.com/2600hz/kazoo5/pull/22
* Required to make `startkey_docid` option work: https://github.com/2600hz/kazoo5/pull/25


kazoo_couch/src/designs/kzc_account_crossbar_listings.erl
```diff
@@ -0,0 +1,237 @@
...
+           }
+     ,<<"device_presence_ids">> =>
+          #{'query_fields' => <<"presence_id">>
+           ,'doc_types' => [<<"device">>]
+           }
+     ,<<"endpoints_lookup">> =>
+          #{'query_fields' => [<<"caller_id_number">>, <<"caller_id_type">>]
+           ,'doc_types' => []
+           }
+     ,<<"faxbox_email_premissions">> =>
```
jamesaimonetti: s/premission/permission

kazoo_documents/src/designs/kzv_faxbox_email_premissions.erl
```diff
@@ -0,0 +1,121 @@
...
+%%% @doc
+%%% @author Hesaam Farhang
+%%%
+%%% This Source Code Form is subject to the terms of the Mozilla Public
+%%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%%
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(kzv_faxbox_email_premissions).
```
jamesaimonetti: s/pre/per

kazoo_couch/src/designs/kzc_account_crossbar_listings.erl
```diff
@@ -0,0 +1,237 @@
...
+%%% * contact_list/names --> crossbar_listings/contact_list_names
+%%% * devices/crossbar_listing --> crossbar_listings/by_type_id
+%%% * devices/listing_by_macaddress --> crossbar_listings/device_mac_addresses
+%%% * devices/listing_by_owner --> crossbar_listings/by_type_ownerid
+%%% * devices/listing_by_presence_id --> crossbar_listings/device_presence_ids
+%%% * devices/sip_credentials --> crossbar_listings/by_sip_credentials
+%%% * directories/crossbar_listing --> crossbar_listings/by_type_id
+%%% * directories/users_listing --> crossbar_listings/owners_by_docid_type
+%%% * faxbox/crossbar_listing --> crossbar_listings/by_type_id
+%%% * faxbox/email_permissions --> crossbar_listings/faxbox_email_premissions
```
jamesaimonetti: s/pre/per

jamesaimonetti (changes requested):
Some `premission` typos should be corrected to `permission`

PR #37, opened 2020-07-15
HELP-15749: start kazoo_number_manager before kazoo_events

jamesaimonetti:
When blackhole runs stand-alone, it relies on kazoo_events for the
kz_hooks when subscriptions for channel events are made. When the
event is a CHANNEL_CREATE from offnet, and no account ID is
associated, kz_hooks_util uses knm_number to lookup the account
associated with the destination.

kazoo_events.app did not list kazoo_number_manager as a dependency;
thus knm was not started and number lookups failed since the knm cache
wasn't started.

Updating kazoo_events.app to its proper dependencies resolves the
issue.


jamesaimonetti:
Helpful test in https://github.com/2600hz/kazoo-properly/pull/8

jamesaimonetti:
Change in https://github.com/2600hz/kazoo-webhooks/pull/4 to not lookup account id either.

lazedo:
@jamesaimonetti i integrated your changes in recent commits to core. i think we can close this. kazoo_events was split, kz_hooks moved to kazoo_hooks

PR #41, opened 2020-07-22
HELP-15121: Fix moh-sound conference setting to work with media-id

harenson:
4.3: https://github.com/2600hz/kazoo/pull/6611


kazoo_media/src/kz_media_util.erl
```diff
@@ -518,6 +522,13 @@ get_prompt(PromptId, Lang, _AccountId, 'false') ->
             kz_binary:join([<<"prompt:/">>, ?KZ_MEDIA_DB, PromptId, Lang], <<"/">>)
     end.
 
+-spec maybe_prompt_path(kz_term:ne_binary(), kz_term:api_ne_binary(), kz_term:api_ne_binary(), {'ok', kz_json:object()} | {'error', 'not_found'}) ->
+          kz_term:api_ne_binary().
+maybe_prompt_path(PromptId, Lang, AccountId, {'error', 'not_found'}) ->
```
lazedo: what about other errors ?

lazedo (changes requested):
needs to handle other errors

kazoo_media/src/kz_media_util.erl
harenson: I also updated `lookup_prompt/1` spec's return value to avoid this happening in the future, thanks :+1:

PR #44, opened 2020-07-24
Removed unused and broken json to csv function. kz_csv:from_jobjs/1 should be used instead.

bradfordben:
Removing unused code as a result of PR https://github.com/2600hz/kazoo-crossbar/pull/29


bradfordben:
Failing test is due to the other half of this pr 2600hz/kazoo-crossbar#29

lazedo:
@bradfordben lets leave the function as-is

bradfordben:
@lazedo Im not sure why you would leave it? It does not work correctly (for json objects with lists) and there is a duplicated function that does the same thing but does work correctly `kz_csv:from_jobjs/1`

Leaving it in place will just result in someone in the future using it and experiencing the same bug/issue around converting json with lists that `kz_csv:from_jobjs` does not have.

bradfordben:
@lazedo Are we sure we want to leave this unused and broken function in the code base when kz_csv:from_jobjs/1 exists and does the same thing but also handles lists inside of JObjs correctly?  I feel it really should be removed and cant see why you want to keep it.

lazedo:
@bradfordben like i said, lets leave it for now. you need of think of callers without lists inside JObj that currently work. what's the rush on deleting this code ?

bradfordben:
@lazedo kz_csv:from_jobjs/1 will handle JObjs with and without lists correctly. 

The only issue i see in the future is someone calling this function and introducing a bug into their code as a result of using this. 

Its also duplicated code as it serves the same purpose as kz_csv:from_jobjs/1.

Im not in a rush to delete it i just cant see the point in keeping it around but if you really dont want to delete it then ok.

lazedo:
@bradfordben merging this `now` will make people having to stop what they're doing and go after the code that still uses it. lets leave it open so we don't forget

bradfordben:
👍

jamesaimonetti:
I checked all our private repos (that I know of anyway :p) and found no usages of the old functionality. Merged

bradfordben:
Thanks @jamesaimonetti

PR #48, opened 2020-07-30
Telnyx carrier module API issues

bradfordben:
- Fixed an issue with the Telnyx number search when searching for tole-free
numbers. When searching for toll-free numbers the api request sent to
telnyx was incorrect resulting in a poor number search response.

- Only set number search options if they are not defined. This will fix a
bug where the quantity of numbers requested during a search was always
overwritten to the max search size that defaults to 500 resulting in
longer query times.

- Fixed broken option "should_keep_best_effort" when searching for numbers
with the telnyx module. The option is now passed to telnyx as part of
the api request as opposed to filtering the response in kazoo.

- Updated broken test telnyx request functions. They were broken as a
result of the changes to the telnyx requests.

Fixes to the Telnyx carrier module

The best effort configuration setting did not work, My guess is that this was written and then telnyx changed the api response and the response to no longer included the flag “any_best_effort” so the old code would never filter the best effort out. This improvement will now pass the best effort flag to telnyx and remove the need to filter the response.

Requests for TF numbers was failing as the request built and passed to Telnyx was incorrect. It passed the TF NAP as prefix in the body and it should be nap as there api doc states. I have also added the option to filer TF numbers by NXX.

Also found that the crossbar quantity option was always overwritten to 500 so every time you requested 15 numbers through CB it requested 500 from telnyx resulting in longer query times. The fix was to only set the default values if they were not set.

4.3 PR https://github.com/2600hz/kazoo/pull/6615


kazoo_numbers/src/knm_search.erl
```diff
@@ -255,11 +255,11 @@ first(Options) ->
...
-    Opts = [{'quantity', ?MAX_SEARCH}
-           ,{'offset', 0}
-           ,{'normalized_prefix', normalized_prefix(Options)}
-            | Options
-           ],
+    Setters = [{'quantity', ?MAX_SEARCH}
+              ,{'offset', 0}
+              ,{'normalized_prefix', normalized_prefix(Options)}
+              ],
+    Opts = lists:foldl(fun props:insert_value/2, Options, Setters),
```
icehess: Can't you use `props:insert_values/2` here?

icehess (changes requested):

kazoo_numbers/src/knm_search.erl
bradfordben: Yes, forgot that one existed, changing it over now

lazedo:
@icehess can you do another review of this one ?

jamesaimonetti:
rebased against latest master. @icehess you have anything more here? lgtm

jamesaimonetti (approved):

bradfordben:
@jamesaimonetti Thanks for rebasing, Looks good to me.

bradfordben:
Applied changes requested on 4.3 equivalent PR

PR #56, opened 2020-08-11
get subprops by key when setting nested proplist

icehess:
Given this expression `props:set_value([key_1,key_sub3], sub3, [{key_1,[{key_sub1,sub_1},{key_sub2,sub2}]}]).`
one would expect `key_sub3` be added to sub-proplist of `key_1`, like this:

```
[{key_1,[{key_sub3,sub3},{key_sub1,sub_1},{key_sub2,sub2}]}]
```

But currently it is replacing the whole `key_1`:

```
[{key_1,[{key_sub3,sub3}]}]
```


jamesaimonetti:
I'm really surprised this works without breaking stuff tbh. Perhaps KAZOO no longer uses string() as a key in any place now? I remember wanting to add this but something was using string() as a key...let me think on it, see if I recall a place

icehess:
I did some grep, so far I couldn't find anything. no matter the key was string or not, I think still the code was wrong. Either it shouldn't handle it or just simply replacing the value at the first level since this was what is happening eventually.

Also there is not a lot of places that actually use `props:set_values`.

PR #57, opened 2020-08-11
generate diff from proposed

lazedo:
* if existing service doesn't have any quantities yet, it crashes.


kazoo_services/src/kz_services.erl
```diff
@@ -1252,7 +1252,7 @@ generate_diff_obj(CurrentServicesJObj, ProposedServicesJObj) ->
     Key = <<"quantities">>,
     PQ = kz_json:get_json_value(Key, ProposedServicesJObj),
     CQ = kz_json:get_json_value(Key, CurrentServicesJObj),
-    Diffs = kz_json:diff(CQ, PQ),
+    Diffs = kz_json:diff(PQ, CQ),
```
k-anderson, line 1255: I am worried that this changes the behavior and is used above on line 1240, feel like we need to dig deeper perhaps into kz_json:diff :)

lazedo: i reverted the commit

PR #63, opened 2020-08-19
[5.0]KZOO-70: use pvt_offering or the literal string and append _reci…

k-anderson:
…pient/_queue to the result

master pr #62


lazedo:
@k-anderson closing this as we still in the phase of merging down from master

PR #64, opened 2020-08-20
Break out math functions to kz_math

jamesaimonetti:
Add clamp/3 to keep value between a lower- and upper-bound


jamesaimonetti:
See also:

https://github.com/2600hz/kazoo5/pull/26

PR #65, opened 2020-08-26
SUPP-14: fix losing origination-call-id when faxing account to account

icehess:
When sending fax, fax_worker sets Origination-Call-ID
but will be discarded by kapi_call on the loopback
and causing `fax_worker` to timeout and retry to the job
again and again.

This will take care of fax events to honor this field.


lazedo:
@icehess this is not needed

icehess:
Closing this as it seems master is not needing this.

PR #75, opened 2020-09-15
PROD-330-Part-1: add QS brackets parser to kz_http_utils

icehess:
Adding a function to parse/normalize the output `cowboy:parse_qs/1` and converts brackets to nested objects/arrays.

This also converts repeated keys in QS to a list or object (e.g. `a=b&a=c => [{<<"a">>,[<<"b">>,<<"c">>]}]`).

The implementation is based on https://github.com/ljharb/qs parser.


lazedo:
@icehess a lot of commented code here, spell checking. why export `merge_objects` ? maybe this part should be move to kz_json ?

kazoo_stdlib/src/kz_json.erl
```diff
@@ -571,6 +574,216 @@ merge_recursive(?JSON_WRAPPER(_)=JObj1, Value, Pred, Keys, Options) when is_func
         'true' -> set_value(Syek, Value, JObj1, Options)
     end.
 
+-define(QSJS_OPTIONS, #{default_empty => 'undefined'}).
+
+-define(IS_JSON(Term), (is_tuple(Term)
```
jamesaimonetti: You cannot rely on this definition of JSON object representation.

Its an extra function call but `kz_json:is_json_term/1` and `kz_json:is_json_object/1` are better to use.

jamesaimonetti (changes requested):

kazoo_stdlib/src/kz_json.erl
icehess: I guess this equivalent to `kz_json:is_json_object/1`:
```
is_json_object(?JSON_WRAPPER(P)) when is_list(P) -> 'true';
```
They both looking for a `P` which a tuple with single element and the element is list.


I need to this reduce clutter and use it in `IS_OBJECT` macro.

icehess: Also we can't use `kz_json:is_json_term/1` here. This is exactly looking for an object. not string, atom, integer or any other values.

jamesaimonetti: Right, today that's how they're defined. Tomorrow it could be `is_json_object(#{}) -> true; is_json_object(_) -> false.` or something. Don't break the abstraction!

PR #77, opened 2020-09-17
Improved kz_json_schema:filter/2 to now support filtering JObjs with the schema's additionalProperties field.

bradfordben:
kz_json_schema:filter/2 was not respecting the schemas “additionalProperties” field when filtering JObjs. As a result if a schema defined an object with “additionalProperties” set to an object then kz_json_schema:filter/2 would all ways filter out the objects.

Eg, if you defined a property of a schema as below and called filter/2 it would all ways be filtered out.

      ``` 
         "new_property": {
            "additionalProperties": {
                "properties": {
                    "additional_property_1": {
                        "description": "Example additional property",
                        "type": "string"
                    }
                },
                "additionalProperties": false,
                "type": "object"
            },
            "description": "Example new property with additional properties",
            "type": "object"
        }

With the new implementation, for it filter out additional properties from a JObj that are not defined in the schemas properties, you must define "additionalProperties": false. This is because by default "additionalProperties" is assumed to be true allowing additional properties to be defined without breaking the schema.

Required schema changes to keep same account config api output : https://github.com/2600hz/kazoo-crossbar/pull/47

4.3 PR: https://github.com/2600hz/kazoo/pull/6631


lazedo (changes requested):
please find another solution

bradfordben:
@lazedo Can you please elaborate on why you do not like this solution? 

As Jesse has a limited number of exported functions its the only way i can correctly filter out keys that do not conform to the schema. I dont want to reimplement all the JSON schema validation logic that Jesse has as that would be crazy when we have a lib to do it.

The current filter function does not support schemas with the "additionalProperties" of type object as the keys are unknown. I cant think of another solution unless you have a better suggestion?

lazedo:
@bradfordben you can start by providing a `existing schema` a `payload` and `what` your calling and `why`you think its not correct. 
changing the existing `schemas` is not what we expect and certainly not for in `4.3`

bradfordben:
> @bradfordben you can start by providing a `existing schema` a `payload` and `what` your calling and `why`you think its not correct.
> changing the existing `schemas` is not what we expect and certainly not for in `4.3`

@lazedo We are trying to add an internal config param called "seat_types" that we will use internally. "Seat_types" would be added to the system_config.accounts and account_config.accounts schema as i will detail below.

```
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "_id": "system_config.accounts",
    "description": "Schema for accounts system_config",
    "properties": {
         ............
        "seat_types": {
            "additionalProperties": {
                "properties": {
                    "display_name": {
                        "description": "The seat type display name",
                        "type": "string"
                    }
                },
                "required": [
                    "display_name"
                ],
                "type": "object"
            },
            "default": {
                "default": {
                    "display_name": "Default Seat Type"
                }
            },
            "description": "List of system wide seat types that can be set on a user",
            "type": "object"
        }
        .........
  }
}
```


```
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "_id": "account_config.accounts",
    "description": "Schema for accounts account_config",
    "properties": {
        "seat_types": {
            "additionalProperties": {
                "properties": {
                    "display_name": {
                        "description": "The seat type display name",
                        "type": "string"
                    }
                },
                "required": [
                    "display_name"
                ],
                "type": "object"
            },
            "description": "List of seat types that can be overlaid on the system level seat types for the account and all descendant accounts",
            "type": "object"
        }
    },
    "type": "object"
}

```

So you will see that as we have defined in the schema the following data is correct: 

```
"seat_types" : {
  "id_1234567" : {
    "display_name": "Seat type A"
  },
 "id_7654321" : {
    "display_name": "Seat type B"
  }
}
```

But if you request the /configs/accounts api endpoint in crossbar you will not receive the seat_types list. This is because the kz_json_schema:filter/2 is too basic to handle the schemas "additionalProperties" attribute. It requires that all the keys are defined in the schema and with the schema defined above the key can be anything (in this case it will be a UUID)

The filter function is the issue as its not respecting the schema correctly so that is why i have addressed this. I understand you might not want to change the schema on 4.3 and thats ok, we dont have to merge it there, but i feel it should be merged in kazoo5.

I know the seat_types is only an internal concept and you wont want it upstreamed but it has just highlighted the filter function issue. The filter function is only called from the config api so this will have no affect on other areas of the code base. 

The need for the schema changes is because by default having additional properties on a json object is valid in the eyes of the schema if "additionalProperties" is not set or is set to "true". So the account config schema is technically not correct (without "additionalProperties" : false) in the way its defined now and the behaviour of the configs api.

bradfordben:
@lazedo Please review my reply when you get a chance. 

@jamesaimonetti Do you have any opinions on this?

bradfordben:
@jamesaimonetti I have rebased off the most reacent master as you requested.

bradfordben:
Thanks @jamesaimonetti

PR #78, opened 2020-09-18
PISTON-1109: Merges the kzd_vmboxes and kzd_voicemail_box modules.

tetrachromium:
The `kzd_vmboxes` and `kzd_voicemail_box` modules both define functions for working with`vmboxes` documents and effectively duplicate functionality. Unlike the `kzd_vmboxes` module, the `kzd_voicemail_box` module has no related `.src` file generated by the document schema. This is inconsistent with other modules for working with database documents.

These changes merge those functions defined by `kzd_voicemail_box` that are actually in use into the `kzd_vmboxes` module. This should not change any functionality, since only one or the other module was actually in use for most cases. The few exceptions have essentially the same implementation in both modules.

Corresponding changes have been made to `core` and the `callflow`, `crossbar`, `properly`, and `teletype` apps. All changes are interdependent and will need to be merged at the same time.

https://github.com/2600hz/kazoo-callflow/pull/13
https://github.com/2600hz/kazoo-crossbar/pull/48
https://github.com/2600hz/kazoo-properly/pull/16
https://github.com/2600hz/kazoo-teletype/pull/11
https://github.com/2600hz/kazoo-core/pull/78


kazoo_documents/src/kzd_box_message.erl
```diff
@@ -309,8 +309,10 @@ metadata(JObj) ->
 
 -spec metadata(doc(), Default) -> doc() | Default.
 metadata(JObj, Default) ->
-    Metadata = kz_json:get_json_value(?KEY_METADATA, JObj, Default),
-    maybe_add_transcription(Metadata, JObj).
+    case kz_json:get_json_value(?KEY_METADATA, JObj, Default) of
```
jamesaimonetti: I think I'd rather add a function clause: `maybe_add_transcription('undefined', _JObj) -> 'undefined';` to handle the result of this `get_json_value` call.

jamesaimonetti (changes requested):
Should also look at adding a run through `scripts/kz_diaspora` to automate changing the various kzd_voicemail_box calls to their kzd_vmboxes equivalents. I don't think private code uses it but just in case!

Thanks for resolving this! Appreciate having help on the janitorial work

kazoo_documents/src/kzd_box_message.erl
tetrachromium: I know nothing about `scripts/kz_diaspora`, but will have a look at it and may get back to you with questions. I just looked into the "check" failures and at least one (xref) seems to be a result of the fact that these changes are interdependent with PRs in other kazoo5 micro-repos. I would expect the same problem with all of this set of related changes.

BTW, this cleanup is some preliminary work for some functional changes I'll be merging into kazoo5 and pushing in the next few days. They'll all have to be based off these branches and are going to encounter the same problems of interdependencies between the micro-repos.

jamesaimonetti: @rpn-voxter yeah, kz_diaspora is a script in the root/scripts folder for managing these movements across all the code. It started with moving kz_util-related functionality out of that god module into more appropriate locations. Its become quite useful in helping migrate other repos' code to use the new calls automatically.

As for the relevant PRs, yes I know the pain of xref in this situation. I'll likely review them all in a clump locally and if all checks pass will merge them as a group.

jamesaimonetti: Also an edoc-related change in kzd_vmboxes to handle

tetrachromium: Can you clarify the edoc-related change in `kzd_vmboxes` please? The only edoc comments are on functions that were moved from `kzd_voicemail_box` without changes. I assumed they would already be correct.

tetrachromium: There is an `applications/crossbar/doc/voicemail.md` file that I think is obsolete, but is still updated when the `vmboxes.json` schema is changed. Would you like me to get rid of this file as part of this PR?

tetrachromium: Apart from my question in the previous comment, I believe everything has been addressed.

PR #91, opened 2020-10-01
PISTON-1134: enforce non-empty CF number when CF is enabled

danielfinke:
Do not allow users to enable call forward if they have entered an empty forwarding number

Related PRs:
2600hz/kazoo-callflow#21


lazedo:
@danielfinke shouldn't this be done in the `schema` ?

danielfinke:
@lazedo is there a way to enforce this in the schema? I tried `"minLength"` on `"number"`, but that still validates if `"number"` is omitted from the `"call_forward"` object.

danielfinke:
@jamesaimonetti since you weighed in on the 4.3 port of this PR, maybe you can also let me know if you have an idea re: the min length

danielfinke:
@jamesaimonetti bump - since the 4.3 version of this has been accepted. Want to know if it should be modified somehow.

PR #100, opened 2020-10-15
KCRO-41: Add end_date field to temporal_rules schema

harenson:
Depends on: https://github.com/2600hz/kazoo-crossbar/pull/65


kazoo_documents/src/kzd_temporal_rules.erl
```diff
@@ -77,6 +78,18 @@ set_enabled(Doc, Enabled) ->
 delete_enabled(Doc) ->
     kz_json:delete_key([<<"enabled">>], Doc).
 
+-spec end_date(doc()) -> integer().
+end_date(Doc) ->
+    end_date(Doc, 6258620160).
```
jamesaimonetti: No magic numbers please. What is this default?

harenson: It was calculated using `start_date`'s default value and adding 1 day to it (86400 seconds). This default value has been already removed for `end_date` in the latest commit.

PR #101, opened 2020-10-16
PISTON-1144: add missing entries for kapi_notifications for voicemail_deleted

danielfinke:
- Causes crashes without these if Preview mode is used for the template via cb_notifications

kazoo-crossbar PR: https://github.com/2600hz/kazoo-crossbar/pull/66


harenson:
@danielfinke please rebase.

PR #106, opened 2020-10-28
KZOO-77: Merge E911 non-address fields

harenson:


kazoo_numbers/src/providers/knm_dash_e911.erl
```diff
@@ -96,7 +96,8 @@ maybe_update_e911(PN) ->
             PN;
         'false' ->
             lager:debug("information has been changed: ~s", [kz_json:encode(E911)]),
-            NewE911 = maybe_update_e911(PN, E911),
+            %% kz_json:merge/1 invoked to update "non address" fields like `notification_contact_emails'
```
jamesaimonetti: It is arity-2 :p

jamesaimonetti:
What about knm_telnyx_911?

jamesaimonetti:
Also can we see a test call to confirm expected inputs + outputs?

kazoo_numbers/src/providers/knm_dash_e911.erl
harenson: :D

harenson:
knm_telnyx_911 uses a different approach, it doesn't merge old and new e911 objects, it just uses the new e911 object as the resulting object.

PR #110, opened 2020-11-03
KCRO-29: add kapi for kapi_registrations:search_req

icehess:


lazedo:
@icehess looks good so far, but it how do you handle the `partial_resp` for `detail` requests ?

icehess:
> @icehess looks good so far, but it how do you handle the `partial_resp` for `detail` requests ?

Isn't kamailio terminates partial response with a `search_resp` message?

lazedo:
but you need to collect them

lazedo:
@icehess check this for reference https://github.com/2600hz/kazoo-crossbar/blob/master/src/modules/cb_presence.erl#L210-L230

PR #123, opened 2020-11-20
KCRO-43: Added option to transition state from aging and available to port_in

holy-batman:
As per 4.3 we were allowed to transition state from available or aging to port_in, use case being a customer ports out and the number is deleted and moved to aging or available state. Customer changes his mind and wants to come back to previous carrier and we want to move the state to port_in.


icehess:
@romanats Port in/out were the very first and very old attempt of having porting number in kazoo. They are replace with port request as far as i know. We don't actually use them anymore (I rarely heard someone using it, but this was ages ago pre 4.3 and 5 for sure).

**No number can be transition to `port_in`**. You have to **create** a number in `port_in` state. https://github.com/2600hz/kazoo-core/blob/master/kazoo_numbers/doc/number_states.md#port_in

As far I know we don't use `port_out` any where as far as i know. This for the old concept. Not sure what we do for now. Maybe hard delete? idk.

icehess:
Basically when a released or deleted or what you called port out then they have to create a new port request.
Look at the code for `cb_port_request:check_number_existence/3` and `knm_numbers:fetch_account_from_number/1`.

icehess:
Those port in/out was the manual port request back in time. They maybe some very old users still using them, that is why their still in the system.

holy-batman:
That makes sense. I have been going on this the wrong way. Will do some tests with porting to make sure we can still do porting when number is available and will ask testing team to add new tests for those.

PR #124, opened 2020-11-23
KZOO-76: added support for IBM speech to text

k-anderson:
4.3 PR: https://github.com/2600hz/kazoo/pull/6669


k-anderson:
@jamesaimonetti thanks, I was looking into dializer between things and starting to scratch my head :)

PR #135, opened 2020-11-28
Strip port from realm in auth request

holy-batman:
Fix issue where realm is not being found when request realm contains port in it.

e.g. 
realm.com:5060


pklinus:
Defect Submitted -> https://2600hz-commercial.atlassian.net/browse/KZOO-85

PR #145, opened 2020-11-30
Rate option default should be empty list

holy-batman:
hon_util:270 uses [] as empty or default value and undefined causes a crash.


kazoo_documents/src/kzd_rates.erl
```diff
@@ -149,7 +149,7 @@ set_iso_country_code(Doc, IsoCountryCode) ->
 
 -spec options(doc()) -> kz_term:api_ne_binaries().
 options(Doc) ->
-    options(Doc, 'undefined').
+    options(Doc, []).
```
jamesaimonetti, line 152: update spec

holy-batman: sorry for dumb question, but what would be the correct return instead of kz_term:api_ne_binaries() is kz_term:api_binaries() good ?

jamesaimonetti: @romanats not dumb! `api_*` in types generally means the type + `'undefined'`. So `api_ne_binaries()` means `ne_binaries() | 'undefined'`. Since the default is now a list, the return can just be `kz_term:ne_binaries()`.

holy-batman: @jamesaimonetti thank you, fixed :)

PR #153, opened 2020-12-03
Add wav support to watson and remove extra code

holy-batman:
Added wav supported and removed unnecessary code which was creating confusion, we already have the preferred audio codec not sure why we are looking for it again.


holy-batman:
https://cloud.ibm.com/docs/speech-to-text-data?topic=speech-to-text-data-audio-formats docs for what content types are supported.

PR #155, opened 2020-12-08
HELP-18604: set inter-server conference extension

jamesaimonetti:


jamesaimonetti:
Will need https://github.com/2600hz/kazoo-crossbar/pull/91

PR #158, opened 2020-12-08
don't default to kz_endpoint_v4 in tests

icehess:
Don't use `kz_endpoint_v4` for tests
Required by https://github.com/2600hz/kazoo-callflow/pull/37


icehess:
ecallmgr tests are failing, reverting to mock kz_endpoint in cf_ring_group_tests.

PR #164, opened 2020-12-11
KCRO-68: change not_reconcilable error code to 400 instead of 404

holy-batman:
when user is sending invalid data and we are sending issue a not_reconcilable it means we its a client issue and error should be 400 instead of 404 not found.


holy-batman:
requires https://github.com/2600hz/kazoo-properly/pull/41/files

PR #172, opened 2020-12-15
KZOO-103: add smart formatting to remove hestiations

holy-batman:
https://cloud.ibm.com/docs/speech-to-text?topic=speech-to-text-output#smart_formatting

Smart formatting remove hesitation as well as formats text more elegantly.


holy-batman:
depends on https://github.com/2600hz/kazoo-crossbar/pull/106

PR #174, opened 2020-12-16
Clean up AMQP API schemas

jamesaimonetti:


lazedo:
@jamesaimonetti tried to rebase and add copyright year, no luck.
why the changes for `Event-Category` and `Event-Name` ? i think that if we go that road, then we should get the required and optional from `kz_api`

PR #178, opened 2020-12-18
CAG-31: Check for different id in options

jamesaimonetti:
Check using from_user as well

delete accounts version if its lingering


lazedo (changes requested):
lets talk

kazoo_documents/src/kzd_fetch.erl
```diff
@@ -128,6 +128,10 @@ fetch_timestamp_micro(JObj, Default) ->
 fetch_user(JObj) ->
     kz_json:get_ne_binary_value(<<"user">>, JObj).
 
+-spec fetch_from_user(data()) -> kz_term:api_ne_binary().
+fetch_from_user(JObj) ->
+    kz_json:get_ne_binary_value(<<"variable_sip_from_user">>, JObj).
+
```
lazedo: is this still needed ?

kazoo_directory/src/kz_directory_device.erl
```diff
@@ -29,9 +29,24 @@ profile(EndpointId, AccountId) ->
...
+        {'error', 'not_found'} ->
+            maybe_from_options(EndpointId, AccountId, Options);
         Error -> Error
     end.
 
+maybe_from_options(EndpointId, AccountId, Options) ->
+    {Endpoint, Without} = props:take_value('endpoint', Options),
+    case kz_json:is_json_object(Endpoint) of
+        'false' -> {'error', 'not_found'};
+        'true' -> maybe_from_options(EndpointId, AccountId, Without, kz_doc:id(Endpoint))
```
lazedo: i think we should be using `Endpoint-ID` or `kzd_endpoint:id(Endpoint)`  instead of `kz_doc:id(Endpoint)`

jamesaimonetti: Hmm, I'll check but when I was logging `Endpoint` i needed `_id` to get the new ID

kazoo_directory/src/kz_directory_device.erl
```diff
@@ -29,9 +29,27 @@ profile(EndpointId, AccountId) ->
...
+            lager:info("found endpoint ~s in options for ~s", [NewId, EndpointId]),
+            maybe_from_options(EndpointId, AccountId, Without, NewId)
+    end.
+
+maybe_from_options(EndpointId, _AccountId, _Options, EndpointId) ->
+    {'error', 'not_found'};
+maybe_from_options(_EndpointId, AccountId, Options, DeviceId) ->
+    lager:info("trying device ~s instead of endpoint ~s", [DeviceId, _EndpointId]),
+    profile(DeviceId, AccountId, Options).
+
```
lazedo: this logic seems a duplication (and variant) of `kz_directory_endpoint` . i like the `props:take_value` so it doesn't repeat, maybe add that logic to `kz_directory_endpoint:maybe_from_options` and drop the changes here.
`kz_directory_endpoint:profile/4` will route to proper module and that's where we can add the `metadevice` handling

kazoo_directory/src/kz_directory_endpoint.erl
```diff
@@ -28,15 +28,20 @@ profile(EndpointId, AccountId) ->
...
 maybe_from_options(EndpointId, AccountId, Options) ->
     Endpoint = props:get_value('endpoint', Options),
     case kz_json:is_valid_json_object(Endpoint) of
-        'true' -> profile(EndpointId, AccountId, Options, kz_doc:type(Endpoint));
-        'false' -> {'error', 'not_found'}
+        'false' -> {'error', 'not_found'};
+        'true' ->
+            NewId = kz_json:get_ne_binary_value(<<"endpoint_id">>, Endpoint, EndpointId),
+            lager:info("found endpoint ~s in options for ~s", [NewId, EndpointId]),
+            profile(NewId, AccountId, Options, kz_doc:type(Endpoint))
```
lazedo: see comment in `kz_directory_device`

PR #179, opened 2020-12-18
SUPP-33: ensure update doc on soft deleted

icehess:
open_doc will ignore if pvt_deleted so update_doc will result in conflict again.
Adding option to allow open pvt_deleted doc.

### Related PR
https://github.com/2600hz/kazoo-crossbar/pull/107


harenson (approved):
:+1:

PR #181, opened 2020-12-22
Update core .app.src files as able

jamesaimonetti:
More to come but these were the "easy" wins that didn't result in
dependency cycles


kazoo_auth/src/kazoo_auth.app.src
```diff
@@ -1,7 +1,8 @@
 {application,kazoo_auth,
-             [{applications,[crypto,kazoo_amqp,kazoo_apps,kazoo_caches,
-                             kazoo_data,kazoo_documents,kazoo_stdlib,
-                             kazoo_web,kernel,lager,public_key,stdlib]},
+             [{applications,[crypto,gen_smtp,kazoo_amqp,kazoo_apps,
```
lazedo: `gen_smtp` looks wrong

jamesaimonetti: @lazedo kz_auth_scope uses `binstr:join` :)

lazedo: can we use an alternative ?

jamesaimonetti: done

lazedo:
@jamesaimonetti circleci failing

PR #185, opened 2020-12-28
Only try options endpoint if ID differs

jamesaimonetti:


lazedo:
@jamesaimonetti is this still valid ?

jamesaimonetti:
@lazedo nope, just reworked the tests without this and all's well.

PR #188, opened 2021-01-05
KCAL-22: Add new parameter to exclude years from temporal_rules

harenson:
Related:
https://github.com/2600hz/kazoo-crossbar/pull/117
https://github.com/2600hz/kazoo-callflow/pull/49


icehess:
Circular dependency to 2600hz/kazoo-crossbar#117

icehess (approved):

PR #191, opened 2021-01-12
update log level of any sink in extra_sinks config

jamesaimonetti:


jamesaimonetti:
See https://github.com/2600hz/kazoo5/pull/41

PR #193, opened 2021-01-12
Load storage dataplan directly

jamesaimonetti:
Slow cache purges may cause incorrect plans. This is mostly seen on CI
runs but could impact slower production systems. Since plans are
cached locally for the lifetime of the node, no need to fetch from the
kzs_cache and risk stale data.


harenson:
:+1:

PR #194, opened 2021-01-14
Fix incorrect types

jamesaimonetti:


jamesaimonetti:
Needs https://github.com/2600hz/kazoo-crossbar/pull/118

PR #195, opened 2021-01-14
[master] set formated db name when fetching attachment

icehess:


icehess:
5.0 PR: https://github.com/2600hz/kazoo-core/pull/203

PR #198, opened 2021-01-15
[5.0] KCRO-84: handle empty mac address on device update

icehess:
When device has been update, provisioner_util should check both old and new mac
address (if any) is define before sending update to provisioner.

Master PR: https://github.com/2600hz/kazoo-core/pull/197


holy-batman:
@icehess I guess the code checks are already fixed by your previous pr and now the now the code needs to be merged or rebased to 5.0, could you please confirm that.

icehess:
@holy-batman pr updated.

PR #202, opened 2021-01-18
KZOO-146: Verify alerts are enabled before creating one

harenson:
Without this change every time kazoo tries to build an alert object and alerts are disabled it crashes because it expects the response from `kapps_alerts:create/5` to be `{ok, AlertObj}` and instead it is receiving `{error, disabled}`. Thus, instead of relying on kapps_alert:create/5 always returning {ok, AlertJObj} check if alerts are enabled before trying to create one.


icehess (approved):

icehess:
@harenson Don't forget to back port to 5.0.

harenson:
@icehess thanks for the reminder, here it is https://github.com/2600hz/kazoo-core/pull/204

PR #203, opened 2021-01-19
[5.0] set formated db name when fetching attachment

icehess:


icehess:
Master PR: https://github.com/2600hz/kazoo-core/pull/195

PR #209, opened 2021-02-10
[master] do not attempt to retry store the attachment if file is not exists

icehess:
Do not retry to store the file if file is not exists in media server. Return an
appropriate error instead so the caller can handle it to not retry it too.

### Related PRs
Not attempt to store in fax: https://github.com/2600hz/kazoo-fax/pull/15
https://github.com/2600hz/kazoo-fax/pull/16


lazedo (changes requested):
change `file_not_exists` to `file_not_found`

lazedo (approved):

PR #211, opened 2021-02-11
PISTON-1086:Fixed incorrectly formatted kazoo_document validation system error responses

bradfordben:
4.3 PR https://github.com/2600hz/kazoo/pull/6698


jamesaimonetti:
@bradfordben just curious, what call sites handle these `throw` exceptions?

bradfordben:
> @bradfordben just curious, what call sites handle these `throw` exceptions?

@jamesaimonetti It is caught by the validate function in each of the kzd modules, eg kzd_users line 1025 or kzd_callflows line 174

bradfordben:
Thanks @jamesaimonetti

PR #213, opened 2021-02-15
Consider 202 as success

jamesaimonetti:


jamesaimonetti:
For https://github.com/2600hz/kazoo-properly/pull/47
https://github.com/2600hz/kazoo-crossbar/pull/142

PR #215, opened 2021-02-17
[ADD-feature] start_app/1, stop_app/1, running_apps/0 to  kapps_maintenance

filipevarjao:
This PR is a proposal to enable start_app, stop and check which apps are running throw the [CB API](https://docs.2600hz.com/supported/applications/crossbar/doc/sup/#execute-a-command-arguments) by calling
`{MODULE}/{FUNCTION}/{ARGS}` in kapps_maintenance.

We lock SUP down to just _maintenance modules. I don't know for sure if we want to expose this feature, though.


filipevarjao:
Since we have  kazoo_apps_maintenance start/1 and stop/1, this PR might no longer need, but with that, we can start and stop something but not add/remove it from autoload

PR #216, opened 2021-02-23
KCRO-56: improve kzd_app accessors

icehess:
- add arity 3 to kzd_apps i18n accesors to return custom default value.
- add path_i18n_* functions


icehess:
This has circular dependency on https://github.com/2600hz/kazoo-crossbar/pull/119 for kzd_app `language`.

PR #217, opened 2021-02-24
Remove unnecessary allotment fns

holy-batman:


holy-batman:
depends on https://github.com/2600hz/kazoo-jonny5/pull/10

PR #218, opened 2021-02-24
[5.0] Remove unnecessary allotment fns

holy-batman:
depends on https://github.com/2600hz/kazoo-jonny5/pull/11


jamesaimonetti:
In master but 5.0 is not receiving these.

PR #227, opened 2021-03-01
use error_on_duplicate_key option when decoding JSON objects

icehess:


icehess:
Required PRs:
* https://github.com/2600hz/erlang-jiffy/pull/1
* https://github.com/2600hz/kazoo5/pull/46

kazoo_stdlib/src/kz_json.erl
```diff
@@ -210,7 +211,7 @@ decode(JSON, Options) ->
     try unsafe_decode(JSON, Options)
     catch
         _:{'invalid_json', {'error', {_Loc, _Msg}}, _JSON} ->
-            lager:error_unsafe("decode error ~s near char # ~b => ~s", [_Msg, _Loc, binary:part(JSON, _Loc, 25)]),
+            lager:error_unsafe("decode error ~s near char # ~b => ~s", [_Msg, _Loc, kz_binary:truncate(_JSON, 25)]),
```
jamesaimonetti, line 219: If the JSON is big and the error is at the end, what use is there in printing the beginning? Perhaps clamping the size to print based on `_Loc - byte_size(_JSON)`?

icehess: I can say same thing for this: if the JSON is whatever size and the error is maybe at start or end (the whole object in this case) what use is there to just printing only nothing? In case of duplicate key here, _Loc is the size of _JSON so `_Loc - byte_size(_JSON)` is `0` so it prints nothing at all.

jamesaimonetti: I would say return the location of the second instance of the key if possible

lazedo: maybe we're looking in the wrong place need to handle better the jiffy exception.
`throw({'invalid_json', {'error', {0, 'decoder_exit'}}, JSON})`

i would leave this without change and check the `_Error:_Reason`  in unsafe_decode

icehess: @jamesaimonetti jiffy `decode.c` is reading the whole object and when is sees `}` then it would pass the whole object as eterm to function to make object. so it doesn't have really the location of keys anymore. changing the `decoder.c` is quite more difficult to keep track of keys.

What I can do it either print the whole JSON. Or change is jiffy to return the name duplicated key (this could be difficult too) Or just print a few chars around the `Loc`.

icehess: @lazedo @jamesaimonetti any thought on how to proceed?

jamesaimonetti: @icehess yeah, no worries, this is fine.

PR #235, opened 2021-03-16
[5.0] disable a test in duo that verifies valid sig_auth because it is expired now

icehess:
Because the test sig_auth is expired, disabling the test until we find another valid one.


icehess:
master: https://github.com/2600hz/kazoo-core/pull/234

PR #238, opened 2021-03-18
Kcro 94 5.0

harenson:
Master: https://github.com/2600hz/kazoo-core/pull/237

For cdrs with version 1 and 2 we needed to subtract the duration seconds from the call's timestamp because they were using the channel destroy as the timestamp value but this is not required anymore on 5.x because cdrs now have channel_created_time field.


harenson:
Wrong destination branch, closing this PR.

PR #243, opened 2021-03-26
[5.0] KAZOO-172: prevent duplicate recordings 

swysor:
Fixes an issue where recordings initiated on the account level were invoked from multiple locations, resulting in duplicate recordings. 

The fix adds an object to the kazoo_call record which tracks recordings by `direction` and `call_id` and can be used to dedupe recordings for all four scenarios in the matrix,

`inbound -> onnet`
`inbound -> offnet`
`outbound -> onnet`
`outbound -> offnet`

and enforces one recording per leg.


lazedo (commented):
i think the fix is in the wrong place. have you looked in `callflow` ?

PR #248, opened 2021-04-01
[5.0] KZOO-173: Strict check ipv4 addresses

holy-batman:
when a number is passed through ip check it results in saying valid ip when its just a number, strict checking of ipv4 will make sure when an ip is properly formatted with dots is accepted as a valid IP.


jamesaimonetti:
@holy-batman properly test to show this? :)

PR #251, opened 2021-04-12
KCRO-97: Add function for building metadata for given object

harenson:
Dependency for:
- https://github.com/2600hz/kazoo-crossbar/pull/175
- https://github.com/2600hz/kazoo-properly/pull/53


kazoo_documents/src/kz_doc.erl
```diff
@@ -790,6 +791,29 @@ remove_pvt(<<"_", Key/binary>>) -> Key;
 remove_pvt(<<"pvt_", Key/binary>>) -> Key;
 remove_pvt(Key) -> Key.
 
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec metadata(doc()) -> kz_json:object().
+metadata(Doc) ->
+    metadata(id(Doc), created(Doc), modified(Doc), kz_json:get_json_value(<<"_read_only">>, Doc)).
```
jamesaimonetti: _read_only will never exist on the doc, it is added to response payloads and stripped if included on requests.

jamesaimonetti (changes requested):

kazoo_documents/src/kz_doc.erl
harenson: It is an special case which is explained within metadata/3, it only  happens with phone_numbers API responses:

```
%% phone_numbers API is the only one that doesn't use crossbar_doc module for loading
%% database documents, it loads those in its own way and the returned document has the
%% created and modified fields within a _read_only object.
```

kazoo_numbers/src/knm_phone_number.erl
```diff
@@ -444,7 +444,8 @@ to_public_json(PN) ->
                ,Features
                ]),
     Root = kz_json:set_values(Values, kz_doc:public_fields(JObj)),
-    kz_json:set_value(<<"_read_only">>, ReadOnly, Root).
+    ExistingRO = kz_json:get_json_value(<<"_read_only">>, JObj, kz_json:new()),
```
jamesaimonetti: This is not a phone numbers thing. `_read_only` is for the API response.

jamesaimonetti (changes requested):

kazoo_numbers/src/knm_phone_number.erl
harenson: ReadOnly was already being handled by this module on this function because numbers response is built on this module, not on the API module (cb_phone_numbers).

jamesaimonetti: Doesn't cb_phone_numbers use crossbar_doc:handle_datamgr_success after fetching the number now? https://github.com/2600hz/kazoo-crossbar/pull/175/files#diff-ceed15a72fa876b80f5d29c69a601b81ded5c873e21ff697d0fff4fbf5ceca49R481

harenson: Yes, but it doesn't have the phone_number document to retrieve the metadata information from. The only data crossbar_doc have access to is the data built by `knm_phone_number:to_public_json/1` function.

PR #254, opened 2021-04-13
[5.0] KZOO-179: add 988 as emergency classifier

holy-batman:


lazedo:
@jamesaimonetti hey, just want to make sure that `classifiers` is supposed to be added as it wasn't there before

PR #257, opened 2021-04-23
PROD-169: Split out failover number from forward number

icehess:
Duplicate call forwarding settings into a failover object (account, user, device) and remove ‘failover’ as a call forward option.

The intend is to allow users to be able to change their call forwarding settings without impacting the failover settings and have failover set independently from forward.


icehess:
Circular Required PR:
https://github.com/2600hz/kazoo-crossbar/pull/179
https://github.com/2600hz/kazoo-ecallmgr/pull/88

Related PRs:
https://github.com/2600hz/kazoo-callflow/pull/70
https://github.com/2600hz/kazoo-callthru-us/pull/16
https://github.com/2600hz/kazoo-locations/pull/8

icehess:
It seems `kz_endpoint_v4` needs especial handling for failover:
https://github.com/2600hz/kazoo-core/blob/master/kazoo_endpoint/src/kz_endpoint_v4.erl#L1286

`maybe_build_failover/2` is calling `create_call_fwd_endpoint/3` which it reads all fwd values from `call_forward` object instead of `call_failover`.

kazoo_endpoint/src/kz_endpoint_v4.erl
```diff
@@ -345,13 +346,24 @@ merge_attribute(<<"name">> = Key, Account, Endpoint, Owner) ->
                                ,kzd_accounts:name(Account)
                                ),
     kz_json:set_value(Key, Name, Endpoint);
-merge_attribute(<<"call_forward">> = Key, _Account, Endpoint, Owner) ->
-    case kzd_devices:call_forward_enabled(Endpoint) of
+merge_attribute(<<"call_failover">> = Key, Account, Endpoint, Owner) ->
+    case kzd_call_failover:enabled(kzd_devices:call_failover(Endpoint)) of
```
lazedo: `kzd_devices:call_failover(Endpoint)` may return `undefined`

kazoo_endpoint/src/kz_endpoint_v4.erl
```diff
@@ -345,13 +346,24 @@ merge_attribute(<<"name">> = Key, Account, Endpoint, Owner) ->
...
         'true' -> Endpoint;
         'false' ->
+            AccountAttr = kzd_accounts:call_failover(Key, Account, kz_json:new()),
+            OwnerAttr = kzd_users:call_failover(Owner, kz_json:new()),
+            EndpointAttr = kzd_devices:call_failover(Endpoint, kz_json:new()),
+            Merged = kz_json:merge([AccountAttr, EndpointAttr, OwnerAttr]),
+            kz_json:set_value(Key, Merged, Endpoint)
+    end;
+merge_attribute(<<"call_forward">> = Key, Account, Endpoint, Owner) ->
+    case kzd_call_forward:enabled(kzd_devices:call_forward(Endpoint)) of
```
lazedo: `kzd_devices:call_forward(Endpoint)` may return `undefined`

lazedo (changes requested):
minor potential crash, the rest looks good

kazoo_endpoint/src/kz_endpoint_v5.erl
```diff
@@ -369,6 +370,16 @@ merge_attribute(<<"name">> = Key, Account, Endpoint, Owner) ->
...
     kz_json:set_value(Key, Name, Endpoint);
+merge_attribute(<<"call_failover">> = Key, Account, Endpoint, Owner) ->
+    EndpointAttr = kz_json:get_ne_value(Key, Endpoint, kz_json:new()),
+    case kz_json:is_true(<<"enabled">>, EndpointAttr) of
+        'true' -> Endpoint;
+        'false' ->
+            AccountAttr = kz_json:get_ne_value(Key, Account, kz_json:new()),
+            OwnerAttr = kz_json:get_ne_value(Key, Owner, kz_json:new()),
+            Merged = kz_json:merge([AccountAttr, EndpointAttr, OwnerAttr]),
+            kz_json:set_value(Key, Merged, Endpoint)
```
lazedo: if the result of the `merge` is empty or `not enabled` then we should `not` set the key. please change this in `call_forward` too

lazedo (changes requested):
just noticed this. please check the result of the merge

kazoo_endpoint/src/kz_endpoint_v5.erl
icehess: Done. 
FYI we always set the value without checking if the merge result is empty or not for all other attributes.

icehess:
@lazedo I also added a migration command to split failover from call_forward.

lazedo:
> It seems `kz_endpoint_v4` needs especial handling for failover:
> https://github.com/2600hz/kazoo-core/blob/master/kazoo_endpoint/src/kz_endpoint_v4.erl#L1286
> 
> `maybe_build_failover/2` is calling `create_call_fwd_endpoint/3` which it reads all fwd values from `call_forward` object instead of `call_failover`.

i would prefer to drop `all` changes from `v4` . `v4` is used to build the endpoint and is not handled in freeswitch as a `kazoo-endpoint`

PR #258, opened 2021-04-28
MNUM-4: fix paginated number queries

jamesaimonetti:
A previous commit (30f7f4dd8fe4ee240b5fc9832b0ca57e26f1f4a5) stopped
overriding the `quantity` in Options with the system configured max
quantity. This resulting in effectively loading only 1 "page" of
results into the cache. Subsequent queries from the API for the next
"page" would then get no results.

This commit restores the "max quantity" setting being put in the
options for carrier queries.

Cleanup of the module includes spawning carrier number search with
monitors and using DOWN messages to indicate when to move on from
waiting for a carrier module to complete, instead of hitting the
search timeout (typically 5s). So if configured to search 5 carrier
modules, the total query time if all 5 spawned processes died would be
25s. Even one carrier module crashing would inflict an unnecessary
delay.


jamesaimonetti:
data point on refactor of monitoring carrier search PIDs:
```
search of knm_carrier1 took 138ms
search of knm_carrier2 took 154ms
search of knm_carrier3 took 154ms
search of knm_carrier4 took 155ms
search of knm_carrier5 took 155ms

search results available after 231ms
```

Previously, knm_carrier4 would cause inconsistent timeouts in the waiting process, so total search time could be 4-5s instead of 200ms.

PR #265, opened 2021-06-07
include authz id/type to find dialplans if existing

jamesaimonetti:


kazoo_numbers/src/converters/knm_converter_regex.erl
```diff
@@ -38,23 +41,36 @@
 
 -define(KEY_E164_CONVERTERS, <<"e164_converters">>).
 
+-spec default_converters() -> kz_json:object().
+default_converters() ->
+    ?DEFAULT_E164_CONVERTERS.
```
jamesaimonetti, line 45: Ignore for now

kazoo_numbers/src/converters/knm_converter_regex.erl
```diff
@@ -28,6 +30,7 @@
           [{<<"^(\\+?1)?([2-9][0-9]{2}[2-9][0-9]{6})", ?DOLLAR_SIGN>>, [{<<"prefix">>, <<"+1">>}]}
           ,{<<"^011(\\d{5,})$|^00(\\d{5,})", ?DOLLAR_SIGN>>, [{<<"prefix">>, <<"+">>}]}
           ,{<<"^[2-9]\\d{7,}", ?DOLLAR_SIGN>>, [{<<"prefix">>, <<"+">>}]}
+          ,{<<"^077(\\d+)", ?DOLLAR_SIGN>>, [{<<"prefix">>, <<"+4477">>}]}
```
jamesaimonetti: Ignore

jamesaimonetti:
@lazedo dialplans are not applied prior to searching for callflows. This is a fix (I believe) to apply the dialplan to the request number, then normalize to e164, then search in callflow patterns.

The usecase appears to be creating a callflow with `"patterns": ["^(\\+447735\\d+)$"]` and a dial_plan on the calling device of `"dial_plan": {"^(35835258)$": {"prefix": "077","description": ""}}` will fail to match the callflow pattern as the dialplan isn't applied prior to normalizing.

Thus when the device dials `35835258`, callflows searches for `+35835258` and not `+447735835258`. However, when dialing `07735835258`, normalization works as expected and the callflow pattern is matched.

We can chat more in my morning but wanted to get this on your radar.

kazoo_call/src/kapps_call.erl
```diff
@@ -1721,3 +1745,29 @@ eq([V1|_], [V2|_], Pos) ->
...
+    case kz_json:get_ne_binary_value([<<"Custom-Channel-Vars">>, CCVKey], RouteReq) of
+        'undefined' -> Id;
+        Contact -> maybe_find_endpoint_id_from_uri(Contact, Id)
+    end.
+
+maybe_find_endpoint_id_from_uri(Contact, Id) ->
+    case kzsip_uri:uris(Contact) of
+        [#uri{ext_opts=Opts1, opts=Opts2}] ->
+            case props:get_value(<<"endpoint_id">>, Opts1 ++ Opts2) of
+                Id -> Id;
```
lazedo, line 1768:
```suggestion
                'undefined' -> Id;
                Id -> Id;
```

PR #269, opened 2021-06-21
PISTON-1289: create caller id if undefined

bradfordben:
Added new config attribute to disable kazoo from auto creating an users caller id name from the first and last name if no
caller id name was set. The default value is set to true so that it wont
change current behavior. This allows kazoo to use the UE caller id if
its set on the call leg.

Other required prs: https://github.com/2600hz/kazoo-crossbar/pull/190 

Kazoo 4.3: https://github.com/2600hz/kazoo/pull/6722


lazedo:
@bradfordben i think this doesn't apply to 5.x

jamesaimonetti:
Cleaning up, please re-open if this is still an issue.

bradfordben:
No problem, Thanks @jamesaimonetti

PR #270, opened 2021-06-21
KZOO-158: Add rename_and_refresh_app functionality

harenson:
After renaming an application's folder, this functionality can be used to rename the application on all the accounts using it.


kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -84,6 +86,114 @@ refresh_app(AppPath, AppUrl) ->
...
+%%------------------------------------------------------------------------------
+%% @doc Document: services/{ACCOUNT_ID}:
+%%      Keys to modify:
+%%      - quantities -> (account | cascade | manual) -> user_apps -> {APP_NAME}
+%%      - quantities -> (account | cascade | manual) -> account_apps -> {APP_NAME}
+%% @end
+%%------------------------------------------------------------------------------
+-spec rename_app_on_services_accountid_doc(kz_term:api_binary(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary()) -> 'ok'.
+rename_app_on_services_accountid_doc(_AppId, PrevName, NewName, Account) ->
+    Keys = [{[<<"quantities">>, Qty, Level, PrevName], [<<"quantities">>, Qty, Level, NewName]}
```
icehess: Do not change service doc at all.

First let change name appear normally on the next bill/or service notification.
Secondly change service plan docs. change the `applications` on the root of document. Then maybe run reconcile  services to update the service doc.

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -84,6 +86,114 @@ refresh_app(AppPath, AppUrl) ->
...
+%%------------------------------------------------------------------------------
+-spec rename_and_refresh_app(file:filename_all(), kz_term:ne_binary()) -> 'ok'.
+rename_and_refresh_app(AppPath, PrevName) ->
+    rename_and_refresh_app(AppPath, PrevName, 'undefined').
+
+-spec rename_and_refresh_app(file:filename_all(), kz_term:ne_binary(), kz_term:api_binary()) -> 'ok'.
+rename_and_refresh_app(AppPath, PrevName, AppUrl) ->
+    ?SUP_LOG_INFO("renaming and refreshing ~s app from ~s~n", [PrevName, AppPath]),
+    {'ok', MasterAccountDb} = kapps_util:get_master_account_db(),
+    {'ok', MetaData} = find_metadata(AppPath),
```
icehess, line 106: Handle erros gracefully. If operation on one account fails the whole operation should not fail. Just print this account failed because of this error

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -84,6 +86,114 @@ refresh_app(AppPath, AppUrl) ->
...
+
+%%------------------------------------------------------------------------------
+%% @doc Document: {ACCOUNT_DB}/apps_store
+%%      Keys to modify:
+%%      - apps -> {APP_ID} -> name
+%% @end
+%%------------------------------------------------------------------------------
+-spec rename_app_on_accountdb_apps_store_doc(kz_term:api_binary(), kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary()) -> 'ok'.
+rename_app_on_accountdb_apps_store_doc(AppId, _PrevName, NewName, Account) ->
+    {'ok', AppDoc} = kzd_apps_store:fetch(Account),
```
icehess: Handle errors gracefully. see above.

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -84,6 +86,114 @@ refresh_app(AppPath, AppUrl) ->
...
+    %% a given customer it has not been renamed yet. First, make sure the application was already
+    %% renamed on its own document (the application document) and the master account DB.
+    rename_app(AppId, PrevName, NewName, [AccountId]);
+rename_app(AppId, PrevName, NewName, [Account | Accounts]) ->
+    ?SUP_LOG_DEBUG("renaming ~s -> ~s for account ~s. ~p accounts remaining",
+                   [PrevName, NewName, Account, length(Accounts)]),
+    lists:foreach(fun(F) -> F(AppId, PrevName, NewName, Account) end
+                 ,[fun rename_app_on_accountdb_apps_store_doc/4
+                  ,fun rename_app_on_services_accountid_doc/4
+                  ]
```
icehess: What about `whitelabel` document?

icehess (changes requested):

kazoo_documents/src/kazoo_documents_maintenance.erl
icehess: Service doc doesn't have `applications`, _service plans_ doc has. For this to work I guess it is better to start from top level account, change all its service plans docs, run `kz_services:reconcile/1` and go a level down and repeat.

That is just a thought. I open to other suggestion :)

Basically if account is sub, you need to update service plans on its reseller or master or wherever that plans is defined. then you need to run reconcile.

There is a change to may skip reconcile, if we change the plan service doc will fix it self over time or on the next monthly run eventually.

kazoo_documents/src/kazoo_documents_maintenance.erl
harenson: At this point it is still doing changes on the master account, if there is any failure at this point, the execution should not continue, just fail.

kazoo_documents/src/kazoo_documents_maintenance.erl
harenson: There is nothing related to applications' metadata within that document, or at least I didn't see anything related.

kazoo_documents/src/kazoo_documents_maintenance.erl
harenson: ServicePlan documents don't have application name anymore, it just makes use of the application-id to keep track of applications enabled for a given accounts. Also, just by renaming the application within the master account, rename the application within apps_store document for each account, and then run `kz_services:reconcile/1` is enough, not need to touch service_plan documents.

PR #275, opened 2021-06-29
[master] KCRO-93: Sharpen query_channels resp validate fun

harenson:
After https://github.com/2600hz/kazoo-crossbar/pull/146 PR change, every time a user performs a `GET /channels/{UUID}` request, there is a chance that this requests gets and 404 response even though the channel exist. It happens because the reply of an ecallmgr from another zone (different than the one handling the call) comes first and that response is passing the validation function used to validate the response because it is only checking if there is a `Channels` key defined on the response and not looking at the value for that response. So, if the response comes from a "different" zone, it is an empty json object. If the response comes from the zone handling the call it contains the requested channel's information.


lazedo:
@harenson this is the wrong place to fix. close this ticket and provide a callback to check the return value in cb_channels instead

harenson:
@lazedo Done, here it is https://github.com/2600hz/kazoo-crossbar/pull/196

PR #277, opened 2021-06-30
PISTON-1290 added function maybe_skip_feature_calls_recording

alexey-khabulyak:
added function maybe_skip_feature_calls_recording  which checks 'record_feature_code_calls' param.
related to https://github.com/2600hz/kazoo-crossbar/pull/194
4.3 PR: https://github.com/2600hz/kazoo/pull/6723


jamesaimonetti:
@alexey-khabulyak hi! please rebase your branch against latest master and force-push it here for CI to re-run.

Note: you can also run most CI checks with `make ci-codechecks ci-docs` from the root of the project. Run `make changed` to get a listing of files that will be processed to make sure you are checking the relevant files changed.

kazoo_call/src/kapps_call.erl
```diff
@@ -1696,6 +1697,22 @@ get_recordings(Call) ->
...
+%% @doc Check if we should record calls to feature codes.
+%% First, we check if the callee number starts with * and then check
+%% if the `record_feature_code_calls' param is set.
+%% If the call is a feature call and record feature code calls is false,
+%% then we skip the call recording.
+%% @end
+%%------------------------------------------------------------------------------
+-spec maybe_skip_feature_calls_recording(call(), kz_term:api_object()) -> boolean().
+maybe_skip_feature_calls_recording(Call, Data) ->
+    try binary:first(callee_id_number(Call)) == 42
```
jamesaimonetti: magic numbers are not helpful. you can use `$*`. Also, wouldn't checking the `request_user(Call)` be a better option? Folks are likely to dial '*97'; is there a case you have where callee has the feature code and request_user doesn't?

kazoo_call/src/kapps_call.erl
```diff
@@ -1696,6 +1697,22 @@ get_recordings(Call) ->
...
+%%------------------------------------------------------------------------------
+%% @doc Check if we should record calls to feature codes.
+%% First, we check if the callee number starts with * and then check
+%% if the `record_feature_code_calls' param is set.
+%% If the call is a feature call and record feature code calls is false,
+%% then we skip the call recording.
+%% @end
+%%------------------------------------------------------------------------------
+-spec maybe_skip_feature_calls_recording(call(), kz_term:api_object()) -> boolean().
+maybe_skip_feature_calls_recording(Call, Data) ->
```
jamesaimonetti: Nit on naming, but `maybe_*` suggests an action might be taken. Here, only a boolean is returned so I would suggest `should_skip_feature_calls_recording` which suggests the boolean return

kazoo_call/src/kapps_call.erl
alexey-khabulyak: > magic numbers are not helpful. you can use $*
Yes, it's much better. Thanks!

I've changed `callee_id_number` to `request_user`. But I can't see the difference. `request_user` is the user part of RURI header, `callee_id_number` is the user part of TO header. Usually, they are the same.

kazoo_call/src/kapps_call.erl
alexey-khabulyak: Done.

alexey-khabulyak:
Hi @jamesaimonetti. Thanks for the comments. 
I've fixed the code. But `make ci-codechecks ci-docs` fails. Do you know why?

```
:: Check for raw JSON usage

processing raw_json: escript: exception error: undefined function kz_term:to_binary/1
  in function  no_raw_json_escript__escript__1625__335081__974996__2:process_file/2 (/usr/local/src/kazoo5/scripts/no_raw_json.escript, line 19)
  in call from lists:foldl/3 (lists.erl, line 1263)
  in call from no_raw_json_escript__escript__1625__335081__974996__2:main/1 (/usr/local/src/kazoo5/scripts/no_raw_json.escript, line 14)
  in call from escript:run/2 (escript.erl, line 758)
  in call from escript:start/1 (escript.erl, line 277)
  in call from init:start_em/1 
  in call from init:do_boot/3 
make[1]: *** [code_checks] Error 127
make[1]: Leaving directory `/usr/local/src/kazoo5'
make: *** [ci-codechecks] Error 2
```

kazoo_call/src/kapps_call.erl
```diff
@@ -1784,6 +1785,22 @@ get_recordings(Call) ->
...
+%%------------------------------------------------------------------------------
+%% @doc Check if we should record calls to feature codes.
+%% First, we check if the callee number starts with * and then check
+%% if the `record_feature_code_calls' param is set.
+%% If the call is a feature call and record feature code calls is false,
+%% then we skip the call recording.
+%% @end
+%%------------------------------------------------------------------------------
+-spec should_skip_feature_calls_recording(call(), kz_term:api_object()) -> boolean().
+should_skip_feature_calls_recording(Call, Data) ->
```
jamesaimonetti, line 1797: I don't think we need the try/catch here. Something more like:

```erlang
should_skip_feature_calls_recording(Call, Data) ->
    is_feature_call(request_user(Call))
        andalso should_record_feature_code_calls(Data).

is_feature_call(<<"*", _/binary>>) -> 'true';
is_feature_call(_) -> 'false'.

should_record_feature_code_calls(Data) ->
    kz_json:is_true(<<"record_feature_code_calls">>, Data, 'true').
```

To me, the should_skip function now reads, in my head at least, a lot more clearly. The actual decision-making is contained to two smaller functions as well. And it uses pattern-matching vs binary:first + equality.

What do you think?

alexey-khabulyak: @jamesaimonetti pattern-matching instead of `binary:first()` looks great(much better than try/catch. I used it to handle empty `callee_id_number`). I like it, thnx.
But I am not sure about `should_record_feature_code_calls(Data)`.
Do you propose to hide implementation behind a "nice" name? 
Is it code style or best practice? I am asking because I am new here) thank you for the comments.

jamesaimonetti: Generally speaking, @alexey-khabulyak, I prefer more, smaller functions. The project generally tries to do so as well.

For instance, in this case, `should_skip_feature_calls_recording/1` implementation is "easy" to read now - just check if is_feature_call andalso should_record_feature_code_calls. Hmm, should probably `not` that result though. I think a "nice" function name is usually easier to read and reason about than a call to `kz_json:is_true(...)`.

Plus I think the compiler can probably optimize all of this under the hood, so I'd rather write for ease of understanding by the reader. As Joe Armstrong says: "Make it work, then make it beautiful, and only if you have to, make it fast".

alexey-khabulyak: @jamesaimonetti got it. Thnx a lot!

PR #278, opened 2021-07-01
[master] KBLA-6: add new kapi for publishing websocker app_event

icehess:
Adding a new KAPI for publishing app_events from blackhole using `publish_app_event` event.


icehess:
Related PRs:
* https://github.com/2600hz/kazoo-blackhole/pull/14
* https://github.com/2600hz/kazoo-properly/pull/57

kazoo_amqp/src/api/kapi_websockets.erl
```diff
@@ -11,6 +11,11 @@
 
 -export([api_definitions/0, api_definition/1]).
 
+-export([app_event/1
+        ,app_event_v/1
+        ,publish_app_event/1
+        ,publish_app_event/2
+        ]).
```
lazedo, line 18:
```suggestion
-export([event/1
        ,event_v/1
        ,publish/1
        ,publish/2
        ]).
```

in a new `kapi_app`

lazedo:
@icehess this change would great to have in a new `kapi_app` so that any app can use this mechanism, i see a lot of use cases

jamesaimonetti:
JIRA Ticket says closed in favor of diff approach

PR #279, opened 2021-07-01
KZOO-190: Fix wave (wav) voicemail recordings transcription

harenson:


lazedo:
have you tried adding `audio/x-wav` to https://github.com/2600hz/kazoo-core/blob/5.0/kazoo_speech/src/kazoo_asr.erl#L40 ?

harenson:
I did and if I recall correctly I think it worked, but I thought it would be better to fix it here.

lazedo:
the problem is that you're fixing this in attachments and in a single module (probably because the report only mentioned that).

we can go in 3 directions with this PR, all of them drop the attachments change

1 fix the exact problem
 - just add to `kazoo_asr`
 - drop these changes

2 try to fix in kz_mime (incomplete as to_extension called directly will still provide audio/x-wav)
- call `normalize_content_type` from `from_filename` after the call to `from_extension`

3 fix it in Makefile
- drop audio/x-wav in favour of audio/wav in generation (my favourite)

harenson:
@lazedo done, I fixed it in the way you suggested, thanks for reviewing :+1:

kazoo_web/Makefile
```diff
@@ -24,17 +24,18 @@ clean-generated:
 $(GEN_FILE):
 	wget -qO $@ $(GEN_URL)
 
-$(GEN_ERL): $(GEN_FILE) $(GEN_ERL_SRC)
+$(GEN_ERL): clean-generated $(GEN_FILE) $(GEN_ERL_SRC)
```
lazedo: `clean-generated` isn't  this going to generate everytime ?

lazedo: the `GEN_URL` doesn't exist so it is created on first `make`, if you need to test again run `make clean-generated && make`

harenson: Yes, if that rule is not executed before rebuilding the file, it will fail saying the new clause would never match because there is another one a few lines earlier that will always match. If we remove that rule from there we need to make sure that it is run manually at least for this time before re-building the kz_mime.erl file.

Let me know if you want me to remove `clean-generated` from there.

PR #294, opened 2021-08-16
KZOO-195: Follow up recording on xfer if enabled account-wide

harenson:
5.0: https://github.com/2600hz/kazoo-core/pull/322

Right now, if Bob calls Alice, and Alice is configured to record calls, the recording will happen, but then, if Alice parks the call, call recording ends. With this change, if call recording is enabled account-wide, then Kazoo will continue recording even after the call was parked.


kazoo_call/src/kapps_call_recording.erl
```diff
@@ -127,7 +127,6 @@ record_call_command(Data, Call) ->
     MediaDocId = ?MATCH_MODB_PREFIX(kz_term:to_binary(Year), kz_date:pad_month(Month), RecordingId),
     DefaultMediaName = get_media_name(kz_binary:rand_hex(16), Format),
     MediaName = kz_json:get_value(?RECORDING_ID_KEY, Data, DefaultMediaName),
-    FollowTransfer = kapps_call:kvs_fetch('recording_follow_transfer', 'false', Call),
```
lazedo: i think you need to fix where this is set (or not)

kazoo_endpoint/src/kz_account_recording.erl
```diff
@@ -67,6 +67,16 @@ maybe_record_outbound(ToNetwork, _Endpoint, Call, Data) ->
...
     end.
+
+-spec maybe_follow_transfer(kapps_call:call()) -> kapps_call:call().
+maybe_follow_transfer(Call) ->
+    {'ok', Account} = kzd_accounts:fetch(kapps_call:account_id(Call)),
+    Key = [kapps_call:direction(Call), kapps_call:inception_type(Call), <<"enabled">>],
+    ShouldRecord = kz_json:get_boolean_value(Key
+                                            ,kzd_accounts:call_recording_account(Account)
+                                            ,'false'
+                                            ),
```
lazedo: you're repeating the logic for determining if call should be recorded. this is already decided

kazoo_endpoint/src/kz_account_recording.erl
```diff
@@ -67,6 +67,16 @@ maybe_record_outbound(ToNetwork, _Endpoint, Call, Data) ->
         'true' ->
             LabeledData = kz_json:set_value(<<"origin">>, ?ACCOUNT_OUTBOUND_RECORDING_LABEL(ToNetwork), Data),
             {'true'
-            ,kapps_call:start_recording(LabeledData, kapps_call:kvs_store('recording_follow_transfer', 'false', Call))
+            ,kapps_call:start_recording(LabeledData, maybe_follow_transfer(Call))
```
lazedo:
```suggestion
            ,kapps_call:start_recording(LabeledData, follow_transfer(Call))
```

kazoo_endpoint/src/kz_account_recording.erl
```diff
@@ -67,6 +67,16 @@ maybe_record_outbound(ToNetwork, _Endpoint, Call, Data) ->
...
+
+-spec maybe_follow_transfer(kapps_call:call()) -> kapps_call:call().
+maybe_follow_transfer(Call) ->
+    {'ok', Account} = kzd_accounts:fetch(kapps_call:account_id(Call)),
+    Key = [kapps_call:direction(Call), kapps_call:inception_type(Call), <<"enabled">>],
+    ShouldRecord = kz_json:get_boolean_value(Key
+                                            ,kzd_accounts:call_recording_account(Account)
+                                            ,'false'
+                                            ),
+    kapps_call:kvs_store('recording_follow_transfer', ShouldRecord, Call).
```
lazedo:
```suggestion
-spec follow_transfer(kapps_call:call()) -> kapps_call:call().
follow_transfer(Call) ->
    kapps_call:kvs_store('recording_follow_transfer', true, Call).
```

kazoo_endpoint/src/kz_account_recording.erl
harenson: You are correct, I did not notice it 🤦🏾‍♂️

lazedo (approved):

PR #297, opened 2021-08-30
add binary_or_binaries json accessor

lazedo:


kazoo_stdlib/src/kz_json.erl
```diff
@@ -1315,6 +1316,23 @@ get_ne_binaries(Key, JObj, Default) ->
...
+-spec get_ne_binary_or_binaries(get_key(), object() | objects()) -> kz_term:api_ne_binary_or_binaries().
+get_ne_binary_or_binaries(Key, JObj) ->
+    get_ne_binary_or_binaries(Key, JObj, 'undefined').
+
+-spec get_ne_binary_or_binaries(get_key(), object() | objects(), Default) -> Default | kz_term:ne_binary_or_binaries().
+get_ne_binary_or_binaries(Key, JObj, Default) ->
+    case get_value(Key, JObj) of
+        'undefined' -> Default;
+        List when is_list(List) ->
+            case lists:filter(fun kz_term:is_ne_binary/1, List) of
```
jamesaimonetti, line 1328: Should we filter or `lists:all` this? Meaning, should it be an error if a non-binary value is in the list?

lazedo: not opposed to that, just followed the pattern for get_ne_binaries. if we change here then well need to change that one too, right ?

jamesaimonetti: Oh, that's fair. Didn't realize it. Yeah, guess matching behaviour makes sense.

PR #298, opened 2021-08-30
Kcro 108 5.0 add screenpop accessor module

kdamas93:


kazoo_documents/src/kzd_screenpops.erl
```diff
@@ -0,0 +1,199 @@
...
+-module(kzd_screenpops).
+
+-export([new/0]).
+-export([custom_parameters/1, custom_parameters/2, set_custom_parameters/2]).
+-export([custom_parameters_custom_url/1, custom_parameters_custom_url/2, set_custom_parameters_custom_url/2]).
+-export([custom_parameters_link_text/1, custom_parameters_link_text/2, set_custom_parameters_link_text/2]).
+-export([custom_parameters_name/1, custom_parameters_name/2, set_custom_parameters_name/2]).
+-export([custom_parameters_render_name_as_label/1, custom_parameters_render_name_as_label/2, set_custom_parameters_render_name_as_label/2]).
+-export([custom_parameters_request_method/1, custom_parameters_request_method/2, set_custom_parameters_request_method/2]).
+-export([custom_parameters_token/1, custom_parameters_token/2, set_custom_parameters_token/2]).
```
jamesaimonetti: I think this is the old accessor module, when `custom_parameters` was defined as `object` in the schema.

kdamas93: Updated :)

PR #309, opened 2021-10-01
Shared pool of listeners for presence events

jamesaimonetti:


kazoo_events/src/kz_event_listener_sup.erl
```diff
@@ -47,6 +48,21 @@ handle_started(Error) -> Error.
...
+
+event_type_from_pid('undefined') ->
+    'undefined';
+event_type_from_pid([{'registered_name', Name}]) ->
+    event_type(Name).
+
+event_type_from_binary(<<?MODULE_STRING, $_, Event/binary>>) ->
+    kz_term:to_atom(Event);
+event_type_from_binary(_Name) -> 'undefined'.
+
```
lazedo, line 65: this is not needed if we pass `Name` from `start_link` -> `handle_started` -> `start_workers`

PR #318, opened 2021-10-29
[5.0] Fix empty cursor

filipevarjao:
When we query the results the cursor might be empty, we are getting bad_identifier error on the UI


jamesaimonetti:
@filipevarjao can you explain the case where this is causing an issue? @icehess any thoughts on the change?

jamesaimonetti:
And is this an issue on master as well?

kazoo_couch/src/kz_couch_view.erl
```diff
@@ -759,7 +759,7 @@ fetch_more(Cursor, Bookmark, Rows) ->
     case get_results(Server, Db, View, ViewOptions) of
         {'error', 'not_found'} when [] =:= RestDbs ->
             ?MAYBE_LOG("either the db ~s or view ~s does not exists", [Db, View]),
-            {'error', 'not_found'};
+            {'ok', {Cursor, []}};
```
lazedo, line 762: `{'error', 'not_found'}` means the view doesn't exist, i think this is working as expected `here`. maybe the check `when [] =:= RestDbs` is not enough here and needs to look if there are positive results from previous `dbs`

filipevarjao:
ok, I'm trying to list the attempts for webhook endpoint 
`/v2/accounts/{{ ACCOUNT_ID }}/webhooks/{{ ID }}/attempts?page_size=50&created_from={{ TIMESTAMP }}&created_to={{ TIMESTAMP }}&_`
and I'm getting the error 

> cause": "webhooks/attempts_by_hook_listing",
        "message": "bad identifier"
    },
    "error": "404",
    "message": "bad_identifier"

tracking this down, the cursor should just be empty, maybe we are missing a handler for this case after this
  let me know if that makes sense @jamesaimonetti

filipevarjao:
if you all agree on that, I can open a PR to master as well to keep all sync

filipevarjao:
the expected would be to get this 

>  All available data has been loaded. No results.

working fine on sandbox though

filipevarjao:
another option would be adding a better match there https://github.com/2600hz/kazoo-core/blob/master/kazoo_data/src/kz_view_cursor.erl#L474

lazedo:
make sure `webhooks/attempts_by_hook_listing` is created

lazedo:
if the view doesn't exist then you are **hidding** problems ?

PR #319, opened 2021-11-01
KCRO-115: add a field to view result indicating of a user doc has photo

oguerrero913:
added valitation to views
 - crossbar_listing
 - list_by_id
to check whether a user's doccument has or not a photo attached


icehess (changes requested):
`users.json` is the old view, we need to update `account-crossbar_listings.json` view too, this the view we use in `cb_users`.

When changing `account-crossbar_listings.json`, add this logic to all places that is creating `features` field (in `by_type_id`, `by_type_name` and `by_type_ownerid`).

Other than that, LGTM.

icehess (approved):

PR #330, opened 2021-11-09
PISTON-1387 case insensitive search for respheaders

alexey-khabulyak:
Kazoo can’t set proper content-type if links don’t contain “.pdf” and google responds with “content-type” header instead of “Content-Type“. So I made the search case insensitive(according to RFC rfc2616)

> HTTP header fields, which include general-header (section 4.5), request-header (section 5.3), response-header (section 6.2), and entity-header (section 7.1) fields, follow the same generic format as that given in Section 3.1 of RFC 822 [9]. Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.


kazoo_web/src/kz_http_util.erl
```diff
@@ -755,7 +755,7 @@ get_resp_header(RespHeader, RespHeaders) ->
 
 -spec get_resp_header(kz_http:field(), kz_http:headers(), Default) -> kz_http:value() | Default.
 get_resp_header(RespHeader, RespHeaders, Default) ->
-    case lists:keyfind(RespHeader, 1, RespHeaders) of
+    case lists:keyfind(string:to_lower(RespHeader), 1, lists:keymap(fun(Key) -> string:to_lower(Key) end, 1, RespHeaders)) of
```
jamesaimonetti: Wonder if you can just use `props` module:
```erlang
get_resp_header(RespHeader, RespHeaders, Default) ->
    props:get_value(kz_term:to_lower_string(RespHeader)
                   ,[{kz_term:to_lower_string(K), V} || {K, V} <- RespHeaders]
                   ,Default
                   ).
```

alexey-khabulyak: yes. it looks better. Thanks.

jamesaimonetti:
@alexey-khabulyak thanks, backporting to 5.0 here: https://github.com/2600hz/kazoo-core/pull/331

PR #332, opened 2021-11-10
[5.0] Send message to parent gen_listener of consuming status

jamesaimonetti:
When the listener_federator process is consuming, send standardized
message to the parent gen_listener of the change.


jamesaimonetti:
Backport of https://github.com/2600hz/kazoo-core/pull/333

PR #350, opened 2021-12-10
[master] KZOO-212: change bandwidth api url

icehess:
Old bandwidth API URL is shutting down. Using the new URL.


icehess:
4.3 pr: https://github.com/2600hz/kazoo/pull/6747

PR #368, opened 2022-01-19
Handle list of strings as parameters for configuration

harenson:
Nice to have for https://github.com/2600hz/kazoo-registrar/pull/12 changes.


kazoo_apps/src/kapps_config.erl
```diff
@@ -599,6 +602,21 @@ set_boolean(Category, Key, Value) ->
 set_json(Category, Key, Value) ->
     set(Category, Key, kz_json:decode(Value)).
 
+-spec set_list_string(config_category(), config_key(), kz_term:text()) -> set_return().
+set_list_string(Category, Key, Values) ->
+    BinList = [kz_term:to_binary(Value) || Value <- string:split(kz_term:to_list(Values), ",")],
```
jamesaimonetti: We have binary:split/3 (`binary:slpit(Values, <<",">>, ['global'])`) and Values should be a `binary()`

harenson: I also added the `trim_all` function, which in case the provided comma-separated list is an empty binary, it won't save it as `[<<>>]` but as an empty list `[]`.

PR #369, opened 2022-01-21
Return empty list if loading behaviours fails

jamesaimonetti:


kazoo_stdlib/src/kz_module.erl
```diff
@@ -109,7 +109,10 @@ hotload(Module) ->
 -spec behaviours(atom()) -> [atom()].
 behaviours(Module)
   when is_atom(Module) ->
-    props:get_value(behaviour, Module:module_info(attributes)).
+    try props:get_value('behaviour', Module:module_info('attributes'), [])
```
icehess: Why `try/catch`? Isn't caller responsibility to check if the `Module` exists?

jamesaimonetti: yeah probably, was convenient at the time. Will recheck if needed for webhooks

PR #370, opened 2022-01-28
check inception from call-forward

lazedo:
* when multiple forwards occurs, inception from offnet that is forwarded comes into new route_req (loopback-b) as onnet


jamesaimonetti:
@lazedo only thing I can think is if other places in the code check only "onnet" and "offnet" for inception values (vs having a catch-all clause).

lazedo:
@jamesaimonetti i think other places will check in `kapps_call:inception_type/1` which will still return **onnet**/**offnet** depending of inception value (`undefined ` or `not undefined`)

PR #373, opened 2022-02-05
KCAL-33: Add branch_extension getters and setters

harenson:
Depends on: https://github.com/2600hz/kazoo-crossbar/pull/241


kazoo_documents/src/kzd_menus.erl.src
```diff
@@ -62,7 +63,7 @@ set_flags(Doc, Flags) ->
 
 -spec hunt(doc()) -> boolean().
 hunt(Doc) ->
-    hunt(Doc, true).
+    hunt(Doc, false).
```
lazedo: commented in crossbar pr

lazedo (changes requested):
should fixed by regenerating after crossbar pr fix

kazoo_documents/src/kzd_menus.erl.src
harenson: Resolved.

PR #378, opened 2022-02-09
KAZOO-226: kapps_controller should handle external apps dependencies

icehess:
Currently kapps_controller is matching the error with App atom,
but if the app is externally fetched and has another external dependency
it fails here because of we only match with original external App name.

This will address the issue by handling the start external app in another
function by calling binding and try re-start the original app.


kazoo_apps/src/kapps_controller.erl
```diff
@@ -110,26 +110,56 @@ start_app(App) when is_atom(App) ->
...
+          {'error', any()}.
+start_external_app(App) ->
+    StartTime = kz_time:start_time(),
+    case application:ensure_all_started(App) of
+        {'ok', Started}=OK ->
+            lager:info("started ~s in ~pms", [App, kz_time:elapsed_ms(StartTime)]),
+            _ = [kz_nodes_bindings:bind(A) || A <- [App | Started], is_kapp(A)],
+            OK;
+        {'error', {App, {"no such file or directory", DotApp}}}=Error ->
+            lager:info("appex failed to load app ~s (~s) successfully", [App, DotApp]),
```
jamesaimonetti: I don't think we should reference this to app exchange. Just "failed to load app..." should suffice

icehess: hehe, just saw this, yup agreed and i already fixed that

icehess: btw we are going to include appex client in core anyway

PR #381, opened 2022-02-18
[5.0] KCAL-33: Set branch_extension to false by default

harenson:


icehess:
Related PR: https://github.com/2600hz/kazoo-crossbar/pull/247

PR #387, opened 2022-03-25
Format checker for elvis.

lordnull:
This checks for quoted atoms. It is required for pr 2600hz/kazoo5#73 in kazoo5 to be useful.


jamesaimonetti:
@lordnull this should be opened against master. Once/if accepted, we can backport to version branches (5.0, 5.1, etc). But as a general rule, PRs should first be against master.

PR #391, opened 2022-03-30
Added intelequent as a provider for e911.

lordnull:
Other changes to support / enable future development:

* added inteliquent utility module to perhaps ease future inteliquent
integrations.
* Created a common store for region names abrviations. (granted, only canada
and us at the moment, and there could be an argument for not hard-coding
anything)
* pulled out some common code for using idomatic erlang 'result' types.
* pulled out a common pattern of #{ ok := [ OkType ], ko := [ ErrorType] } so
the pattern is easier to discover by future devs.


jamesaimonetti:
please refresh and resubmit if still applicable

PR #401, opened 2022-04-19
KCRO-135: Add 'video' value to summary result for conferences

jamesaimonetti:


jamesaimonetti:
See test in https://github.com/2600hz/kazoo-properly/pull/77

PR #402, opened 2022-04-19
Added portabilty check to knm_gen_carrier.

lordnull:
This also includes updates to make the knm_inteliquent module more
capable.

Futher includes helper data modules for what seems to be a common pattern.


kazoo_numbers/src/carriers/knm_carriers.erl
```diff
@@ -111,6 +112,61 @@ check_numbers(Module, Nums) ->
...
+%%------------------------------------------------------------------------------
+%% @doc Normalize then query the various providers for whether numbers are portable
+%% @end
+%%------------------------------------------------------------------------------
+-spec check_portability(kz_term:ne_binaries()) -> kz_json:object().
+check_portability(Numbers) ->
+    Nums = lists:usort([knm_converters:normalize(Num) || Num <- Numbers]),
+    lager:info("attempting to check portability for ~p", [Nums]),
+    Okkos = lists:map(fun(Carrier) -> check_number_portability(Carrier, Nums) end, available_carriers([])),
+                                                % Now that we have the carriers opnions, we try to sus out the truth.
```
jamesaimonetti: comments on own lines should be `%%`

kazoo_numbers/src/carriers/knm_carriers.erl
```diff
@@ -111,6 +112,61 @@ check_numbers(Module, Nums) ->
...
 
+%%------------------------------------------------------------------------------
+%% @doc Normalize then query the various providers for whether numbers are portable
+%% @end
+%%------------------------------------------------------------------------------
+-spec check_portability(kz_term:ne_binaries()) -> kz_json:object().
+check_portability(Numbers) ->
+    Nums = lists:usort([knm_converters:normalize(Num) || Num <- Numbers]),
+    lager:info("attempting to check portability for ~p", [Nums]),
+    Okkos = lists:map(fun(Carrier) -> check_number_portability(Carrier, Nums) end, available_carriers([])),
```
jamesaimonetti: Generally, using list comprehensions is preferred unless a compelling reason not to.

kazoo_numbers/src/carriers/knm_carriers.erl
```diff
@@ -111,6 +112,61 @@ check_numbers(Module, Nums) ->
...
+                                                % Now that we have the carriers opnions, we try to sus out the truth.
+                                                % the differ promote indicates we want to keep successes, even if other
+                                                % carriers say it's not portable.
+    MergedOkko = kz_okko:merge_keyed(Okkos, kz_okko:build_conflict_resolver([{'differ', 'promote'}])),
+    AsResults = kz_okko:to_results(MergedOkko),
+    InexpliciteResults = lists:foldl(fun({_ResultTag, {PhoneNumber, _}}, NumberSet) ->
+                                             ordsets:del_element(PhoneNumber, NumberSet)
+                                     end, Nums, AsResults),
+    Mapper = fun
```
jamesaimonetti: As this fun doesn't wrap anything from the outer scope, make it a named function and do `Mapper = fun named_fun/1,`

kazoo_numbers/src/carriers/knm_carriers.erl
```diff
@@ -111,6 +112,61 @@ check_numbers(Module, Nums) ->
...
+                                    ,<<"portability_status">> => <<"confirmed">>
+                                    ,<<"not_portable_reason">> => 'null'
+                                    };
+        ({error, {PhoneNumber, Why}}) ->
+                                   #{<<"phone_number">> => PhoneNumber
+                                    ,<<"portable">> =>'false'
+                                    ,<<"not_portable_reason">> => Why
+                                    }
+                           end,
+NumbersResults = lists:map(Mapper, AsResults),
```
jamesaimonetti: Or instead, just `[named_fun(Result) || Result <- AsResults]`

kazoo_numbers/src/carriers/knm_carriers.erl
```diff
@@ -111,6 +112,61 @@ check_numbers(Module, Nums) ->
...
+OutJson = #{ 'inexplicit_result' => InexpliciteResults, 'phone_numbers' => NumbersResults},
```
jamesaimonetti: multi-element lists and maps should drop elements to new lines:

```erlang
#{ 'inexplicit_result' => InexplicitResults
  ,'phone_numbers' => NumbersResults
},
```

jamesaimonetti: Also, creating a map just to convert to JSON seems...dubious.

Just use `kz_json:from_list/1` directly

kazoo_numbers/src/carriers/knm_inteliquent.erl
```diff
@@ -88,6 +132,388 @@ is_number_billable(_Number) -> 'true'.
...
+%%------------------------------------------------------------------------------
+%% @doc Check if the numbers are portable.
+%% @end
+%%------------------------------------------------------------------------------
+-spec check_portability(kz_term:ne_binaries()) -> kz_okko:okko({kz_term:ne_binary(), kz_term:ne_binary()}, {kz_term:ne_binary(), #{}}).
+check_portability(Numbers) ->
+    %% inteliquent wants numbers normalized different from how kazoo normalizes
+    %% them. So, we'll go ahead and do that. But to ensure we provide a valid
+    %% response to kazoo, we need to map from the mutilated numbers.
+    StripNumber = fun(<<"+1", Number/binary>>) -> Number;
```
jamesaimonetti: This should be a named function and called directly in the fold below. I would also put the fold's function in a named function:

```erlang
lists:foldl(fun map_number_to_inteliquent/2, #{}, Numbers),
...

map_number_to_inteliquent(OriginalNumber, InputMap) ->
    maps:put(to_inteliquent(OriginalNumber), OriginalNumber, InputMap).

to_inteliquent(<<"+1", Number/binary>>) -> Number;
etc etc etc
```

kazoo_numbers/src/carriers/knm_telnyx.erl
```diff
@@ -66,6 +67,33 @@ is_number_billable(_PN) -> 'true'.
           {'error', any()}.
 check_numbers(_Numbers) -> {'error', 'not_implemented'}.
 
+%%------------------------------------------------------------------------------
+%% @doc Check with carrier if these numbers are portable
+%% @end
+%%------------------------------------------------------------------------------
+-spec check_portability(kz_term:ne_binaries()) -> kz_okko:okko({kz_term:ne_binary(), kz_term:ne_binary()}, {kz_term:ne_binary(), #{}}).
+check_portability(Numbers) ->
+    Jobj = portability(Numbers),
```
jamesaimonetti: General style thing - `JObj` is preferred vs job-j :)

kazoo_numbers/src/carriers/knm_telnyx.erl
```diff
@@ -66,6 +67,33 @@ is_number_billable(_PN) -> 'true'.
...
+    case kz_json:to_map(Jobj) of
```
jamesaimonetti: I don't think converting to maps to then pattern match for the phone_numbers key/value check below is all that useful, is it?

`case kz_json:get_list_value(<<"phone_numbers">>, JObj) of [_|_] =Raw -> ...` does the test just as well, right?

Everntually kz_json will move to use maps as the internal representation but we should maintain the abstraction.

kazoo_numbers/src/carriers/knm_telnyx.erl
```diff
@@ -66,6 +67,33 @@ is_number_billable(_PN) -> 'true'.
...
+            OkkoFun = fun partition_portabilty_result/1,
+            kz_okko:foldl(OkkoFun, PhoneNumbers);
+        _InvalidObj ->
+            lager:debug("expected a json object with a key of 'phone_numbers' that is a list, not: ~p", [Jobj]),
+            kz_okko:empty_okko()
+    end.
+
+partition_portabilty_result(Jobj) ->
+    PhoneNumber = maps:get(<<"phone_number">>, Jobj),
+    IsPortable = maps:get(<<"portable">>, Jobj, 'false'),
```
jamesaimonetti: Not a change request, but I generally prefer, if the value is only then checked via `case`, to just do the function call in the `case`:

```erlang
case maps:get(<<"portable">>, JObj, 'false') of
    'false' -> ...;
    'true' -> ...
end
```

kazoo_numbers/src/carriers/knm_telnyx.erl
```diff
@@ -219,4 +247,13 @@ search_prefix('npa', NPA, NXX) ->
 -spec should_keep_best_effort() -> {kz_term:ne_binary(), boolean()}.
 should_keep_best_effort() -> {<<"best_effort">>, ?SHOULD_KEEP_BEST_EFFORT}.
 
+-spec portability(kz_term:ne_binaries()) -> kz_json:object().
+portability(Numbers) ->
+    FormattedNumbers = lists:map(
+                         fun(Num) -> kz_json:set_value(<<"phone_number">>, Num, kz_json:new()) end
```
jamesaimonetti: named functions > anon when outer-scope bits aren't needed.

meaning `[set_pn(Num) || Num <- Numbers]` is preferred to lists:map + anonymous functions

kazoo_numbers/src/carriers/knm_telnyx.erl
lordnull: At this point in the code I don't actually know for certain if the json I got back from telnyx is actually an object. The kz_json:get_list_value/2 will crash when given a non-object, while it just spits back what I gave it otherwise. Thus, the conversion to a map so I can leverage erlang's pattern matching to check all that for me.

kazoo_numbers/src/carriers/knm_carriers.erl
lordnull: > Also, creating a map just to convert to JSON seems...dubious.
> 
> Just use `kz_json:from_list/1` directly

The `kz_json:from_map/1` is there pretty much only to satisfy dialyzer since the json objects are encoded there using jiffy's old style structure. I think we should move to using maps to represent json objects.

* It's consistent with the rest of the json terms. All other terms use a json term that can be checked in a guard, and mirrors the json term nearly exactly.
* Using maps removes the 'is this an empty list or an empty object' ambiguity.
* Using maps means we can skip calls to kz_json aside from encoding and decoding due to the above.
* Using maps allows us to pattern match on json directly should we choose to do so.

All the above is why the list of numbers is a list of maps rather than a proplist.

jamesaimonetti: i 100% agree on moving to using maps as the internal representation of JSON objects; however, the undertaking is not as simple as asking jiffy to decode to maps and being done.

I still think kz_json will need to be used for operating on JSON data; there's a lot of hard-won code in kz_json functions, not just getters/setters. The project has already had the pain of converting code that was pattern-matching on `{struct, []}` / mochijson style JSON objects when that changed to `{[]}` instead.

There's a balance to be struck between leveraging pattern-matching for JSON as map vs kz_json handling the internal representation and JSON being considered `opaque` to the rest of KAZOO. I could see a day where OTP does package a JSON codec module that uses a fancy data structure specifically for JSON object at which point all KAZOO code using maps directly would need converting again.

So, I hear you on the upside; we have to weigh that against effort to adapt the existing code base and the potential future versions of JSON representation.

For now, I will ask that `kz_json` and its functions be what's used for JSON manipulation.

kazoo_numbers/src/carriers/knm_telnyx.erl
jamesaimonetti: knm_telnyx_util:rep/1 suggests either a JSON object is returned or `throw` errors.

Is there a response from Telnyx you're getting that returns from telnyx_util:rep/1 that doesn't invoke the `throw` errors but also isn't a JSON object?

kazoo_numbers/src/carriers/knm_carriers.erl
lordnull: I've made the change as it's not a hill worth dying on. I did some things that surprised me:

`kz_json:from_list/1` is not recursive. There is a recursive version, however it expects sublists, and thus clobbered the maps I was using for the sub-objects. Seeing as `kz_json:from_map/1` already existing, and is recursive by default, it struck me as the preferred way to do it (given my viewpoint in my previous comment). In short, the `kz_json:from_map/1` seems to better adhere to the principle of least suprise.

manushiperera:
Hello @lordnull,
This PR has conflicts with the latest master branch. 
Is this still relevant?

jamesaimonetti:
Split these into a portability PR and improvement PR and resubmit if still applicable

PR #403, opened 2022-04-23
[master] KCRO-134: do not check for db_exists when creating ranges

icehess:
Checking for db existence is a side-effect and be better removed.
But more importantly this will potentially returns and empty list.
From KCRO-134 ticket it turns out the missing MODb (with no previous MODB since the account is new)
is causing empty list in crossbar_view which in return falls back and uses AccountDb (which is weird by itself).

Removing this will erase the side-effect aspect of this function. All other places that use this function
can use get_results from this module to properly handle missing db (and by kz_datamgr proxy, missing views)
and get appropriate db_not_found error if they provide the option.

**Update**

By not filtering non existence dbs, crossbar without https://github.com/2600hz/kazoo-core/pull/403 will causing the include previous month for normal modb range, which in case of new accounts or last modb will cause the a legitimate query with some result return error instead.

This PR also intriduce a fix in kz_couch_view will handles the not_found specifically and checks if we have some result from other queried dbs, if not return the last error or continue to next db and return results.


icehess:
5.0: https://github.com/2600hz/kazoo-core/pull/404

PR #405, opened 2022-04-25
Appex client updates

jamesaimonetti:
Includes dialyzer-related updates and an Elvis style checker update
for nested map keys.


kazoo_appex/src/appex_client_client.erl
```diff
@@ -54,15 +54,15 @@
        ).
 -type fetch() :: #fetch{}.
 -type fetches() :: [fetch()].
--type apps() :: #{App::atom() => #{pid => pid()
-                                  ,ref => reference()
-                                  }
+-type apps() :: #{atom() => #{pid => pid()
```
icehess, line 57: Why removing `App`? I like that. It allows you to read what that atom reads.

kazoo_appex/src/appex_client_client.erl
```diff
@@ -784,9 +786,6 @@ link_cluster(ReqData) ->
 process_link_cluster(ReqData, Body) ->
     ShouldRestart = kz_json:is_true(<<"restart_appex_client">>, ReqData),
     case kz_json:get_ne_json_value([<<"data">>, <<"attributes">>], Body) of
-        'null' ->
```
icehess, line 787: Please don't remove this as appex server may respond with `{ "data": null }`

kazoo_appex/src/appex_client_client.erl
```diff
@@ -862,9 +861,6 @@ unlink_cluster() ->
 
 process_unlink_cluster(Body) ->
     case kz_json:get_ne_json_value([<<"data">>, <<"attributes">>], Body) of
-        'null' ->
```
icehess, line 865: same as above appex server may respond with `{ "data": null }`

icehess (commented):
thanks for your review @jamesaimonetti

kazoo_appex/src/appex_client_client.erl
jamesaimonetti:
```erlang
1> kz_json:get_ne_json_value([<<"data">>, <<"attributes">>], {[{<<"data">>, null}]}).
** exception error: bad argument
     in function  kz_json:get_value1/3 (src/kz_json.erl, line 1462)
     in call from kz_json:get_ne_json_value/3 (src/kz_json.erl, line 1072)
```
What am I missing? How does this function call return `null`?

jamesaimonetti:
```erlang
2> kz_json:get_ne_json_value([<<"data">>, <<"attributes">>], {[{<<"data">>, {[{<<"attributes">>, null}]}}]}).
undefined
```

icehess: Oh right, I forgot I changed it to `get_ne_json_value`. `get_value` returns `null`.

icehess: well the problem is that server may respond with `{data: null}` and both get_value and get_ne_json_value can crash

icehess: anyway i'm going to merge this, I don't expect too much this could happen as server either should have set error or set the data as expected.
I saw in earlier days some crashes won't set error, but I guess I handled them since then.

PR #407, opened 2022-04-26
KZOO-237: Anonymize CCVs->FromURI when privacy is enforced

harenson:
Before, when resource's `privacy=full` and `system_config/privacy/method =
kazoo` if an outbound call was placed only the *name* was being anonymized within the **FROM** header within the **INVITE** being sent to callee.


kazoo_endpoint/src/kz_privacy.erl
```diff
@@ -98,14 +98,12 @@ get_default_privacy_mode() ->
 %%------------------------------------------------------------------------------
 -spec should_hide_name(kz_json:object()) -> boolean().
 should_hide_name(JObj) ->
-    Keys = [[<<"caller_id_options">>, <<"outbound_privacy">>]
-           ,[<<"privacy">>, <<"hide_name">>]
+    Keys = [[<<"privacy">>, <<"hide_name">>]
```
jamesaimonetti: This changes the order things are searched for. Now the `caller_id_options.outbound_privacy` key won't take precedence. Is that the expected outcome?

PR #412, opened 2022-05-03
cache fixes for directory/endpoint & others

lazedo:


kazoo_stdlib/src/kz_maps.erl
```diff
@@ -92,6 +93,39 @@ get(NotAList, Value, Default)
     get([NotAList], Value, Default);
 get([], Value, _) -> Value.
 
+-spec put(term(), map(), term()) -> term().
+put(Key, Map, Value) ->
```
jamesaimonetti, line 97: Can you add some tests that show the expected usage of put? Like what the purpose of the `filter` fun is for?

lazedo: sorry, just noticed this. `put` has the same usage of `get`, ie, give list of keys and use `recursion`. `filter` is just to remove the the empty values by default

PR #413, opened 2022-05-03
KZOO-223: convert direction without regarding duration

jamesaimonetti:
Previously, the code expected the rewind duration to be
negative. However, ecallmgr expects duration to be positive and
negates it if the direction is 'rewind'.

Just check that duration is positive in the seek_command header and
convert the atom direction to binary.


jamesaimonetti:
See backports in https://github.com/2600hz/kazoo-core/pull/415 and https://github.com/2600hz/kazoo-core/pull/414

PR #416, opened 2022-05-03
Cleanup kazoo_call with elvis and dialyzer

jamesaimonetti:


kazoo_call/src/kapps_conference.erl
```diff
@@ -573,25 +578,33 @@ build_conference_profile(#kapps_conference{profile_name='undefined'}=Conference)
 build_conference_profile(#kapps_conference{profile_name=ProfileName
                                           ,base_profile_name=BaseProfileName
                                           ,account_id='undefined'
-                                          }=Conference) ->
+                                          }=Conference
+                        ) ->
     lager:debug("no account id for profile ~s, building from defaults", [ProfileName]),
 
-    BaseProfile = profile(Conference#kapps_conference{profile_name=BaseProfileName, profile=undefined}),
+    {BaseProfileName, BaseProfile} = profile(Conference#kapps_conference{profile_name=BaseProfileName
```
jamesaimonetti: @lazedo this seems required, otherwise the kz_json:merge/2 below would crash, right?

lazedo: yes, was about to commit the change :)

PR #417, opened 2022-05-04
conference profile hierarchy & build simplification

lazedo:
* profile has base_profile that has base_profile, ...
  it now processes "video <- test <- test-1" from account merged up to system
  values from right apply on top of left, 'null' values delete the value from left value
  example, color => blue, color => green, color => null makes color = blue

* processes language agnostic and language specific

* converts media url for language for well configured prompts

* adds utility to kapps_account_config to parse path to config key, helpful for use in sup
* exposes fetch_value/2 from kzd_accounts
* adds fetch_tree to kzd_accounts


kazoo_call/src/kapps_conference.erl
```diff
@@ -1131,3 +977,50 @@ extra_settings(#kapps_conference{conference_doc=Doc}) ->
...
+profile_default(?VIDEO_PROFILE_NAME) -> kz_json:from_list(?VIDEO_PROFILE_CONFIG);
+profile_default(_Other) -> kz_json:new().
+
+profile_names(AccountId, ProfileName) ->
+    profile_names(AccountId, ProfileName, []).
+
+profile_names(_AccountId, undefined, Acc) -> Acc;
+profile_names(AccountId, ProfileName, Acc)
+  when is_binary(ProfileName) ->
+    BaseProfile = kapps_account_config:get_global_merged(AccountId, ?CONFERENCE_CONFIG_CAT, [<<"profiles">>, ProfileName, <<"base_profile">>]),
```
jamesaimonetti: BaseProfile would be either `undefined | kz_json:object()` right? So if its an object, won't the next call into profile_names/3 crash?

kazoo_apps/src/kapps_account_config.erl
```diff
@@ -800,3 +803,50 @@ remove_config_setting(AccountDb, [{Id, Node, Setting} | Keys], JObj, Removed) ->
...
 config_setting_key(Node, Setting) ->
     [Node, Setting].
+
+-spec get_global_merged(kz_term:api_ne_binary(), kz_term:ne_binary(), config_key()) -> kz_json:object().
+get_global_merged(AccountId, Category, Key) ->
+    get_global_merged(AccountId, Category, Key, undefined).
+
+-spec get_global_merged(kz_term:api_ne_binary(), kz_term:ne_binary(), config_key(), kz_term:api_object()) -> kz_json:object().
+get_global_merged(undefined, Category, Key, Default) ->
+    get_global_merged_default(Category, key(Key), Default);
```
jamesaimonetti: Since this could return the value in `Default`, can we change the spec to:

`-spec get_global_merged(kz_term:api_ne_binary(), kz_term:ne_binary(), config_key(), Default) -> Default | kz_json:object().`

kazoo_call/src/kapps_conference.erl
lazedo: not coerced but its either `undefined` or `binary`

kazoo_apps/src/kapps_account_config.erl
lazedo: you're the `spec` master, counting on you :)

kazoo_call/src/kapps_conference.erl
lazedo: maybe `move` these new routines in `kapps_account_config`  into `new module` `kapps_merged_config ?`  and add `get_ne_binary/get_json/get_boolean,....` so we can properly request for the correct the correct types, thoughts ?

kazoo_apps/src/kapps_account_config.erl
lazedo: but i think in this case it should be `Default | kz_json:json_term()`

jamesaimonetti: wfm

kazoo_call/src/kapps_conference.erl
jamesaimonetti: i think that makes sense. As-is its hard to know. Maybe the var could also be `BaseProfileName` to suggest binary() too? Since `*Profile` variables have tended to be the JSON objects of the profile's configs.

kazoo_documents/src/kzd_accounts.erl
```diff
@@ -800,6 +801,10 @@ fetch_name(Account) ->
 fetch_realm(Account) ->
     fetch_value(Account, fun realm/1).
 
+-spec fetch_tree(kz_term:api_ne_binary()) -> kz_term:api_ne_binary().
```
jamesaimonetti: should return `kz_term:api_ne_binaries().`

lazedo: removing this as i just noticed that `tree/1` accepts `AccountId`

PR #382, opened 2022-02-22
KZOO-229: adding appex client to core

icehess:

PR #429, opened 2022-05-11
[5.1] KZOO-229: adding appex client to core

icehess:


icehess:
Backport of #382 #405 and #275 

Circular PR: https://github.com/2600hz/kazoo-crossbar/pull/281

PR #434, opened 2022-05-13
fix some issue in appex regarding mui and crossbar and settings

icehess:
Fixing some issue regarding some issues found during testing:

* fix a defect of not setting the source_url for monster ui app.json metadata
* setting cors headers in appex_cb_router so browser loads the assets
* fixing getting a wrong value from kazoo_appex app env, and forcing the proper value in case of undefined
* fixing an issue with restart app causing crossbar get amnesia about its router dispatch :) 
* add appex deps to list of app's deps so they get started too and the app is fully functional


kazoo_appex/src/appex_cb_router.erl
```diff
@@ -45,19 +46,20 @@ start_link() ->
...
-    lager:info("starting up appex app router"),
-
-    [{Host, Thing, Paths}] = persistent_term:get('crossbar_dispatch'),
-
-    persistent_term:put('crossbar_dispatch'
-                       ,[{Host, Thing, [Path || Path={_, _, Mod, _} <- Paths, Mod =:= ?MODULE]}]
-                       ),
-
-    lager:info("removed our path").
+    %% lager:info("starting up appex app router"),
```
jamesaimonetti: Kill dead code, kill!

PR #435, opened 2022-05-14
[5.1] fix some issue in appex regarding mui and crossbar and settings

icehess:
Fixing some issue regarding some issues found during testing:

* fix a defect of not setting the source_url for monster ui app.json metadata
* setting cors headers in appex_cb_router so browser loads the assets
* fixing getting a wrong value from kazoo_appex app env, and forcing the proper value in case of undefined
* fixing an issue with restart app causing crossbar get amnesia about its router dispatch :) 
* add appex deps to list of app's deps so they get started too and the app is fully functional


icehess:
backport of (#434)

PR #439, opened 2022-05-25
amqp restart/reconnect

lazedo:


kazoo_amqp/src/kz_amqp_channel.erl
```diff
@@ -85,14 +89,22 @@ is_consumer_pid_valid() ->
...
--spec consumer_channel() -> pid() | kz_amqp_assignment() | {'error', 'timeout'}.
+-spec consumer_channel() -> pid().
 consumer_channel() ->
     case get('$kz_amqp_consumer_channel') of
         'undefined' ->
             #kz_amqp_assignment{channel=Channel} = kz_amqp_assignments:get_channel(),
             put('$kz_amqp_consumer_channel', Channel),
             Channel;
-        Channel -> Channel
+        Channel ->
```
jamesaimonetti, line 99: I think is_process_alive/1 can be in a guard. Could be:

```
Channel when is_pid(Channel), is_process_alive(Channel) -> Channel;
_Else ->
    #kz_amqp_assignment{...}
end

lazedo: image

PR #448, opened 2022-06-17
[5.1] KQUB-137: update to use new features array

mark2600:


kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -158,7 +158,11 @@ js_quantify_map() ->
        "            emit(['users', doc.priv_level], 1);"
        "            if (doc.qubicle && doc.qubicle.enabled) {"
        "                if (doc.qubicle.recipient) {"
-       "                    emit(['qubicle', (doc.qubicle.recipient.offering || 'basic') + '_recipient'], 1);"
+       "                    if(doc.qubicle.recipient.features == ['pro']){"
```
k-anderson: @mark2600 if features is going to be an array, shouldn't we be checking if the array contains the value 'pro' rather than if the array is a single element of 'pro'?

mark2600: yes it is, i was not sure how much JS is too much JS for the views

PR #450, opened 2022-06-25
KZOO-265: Maybe hide From-URI when enforcing privacy

harenson:
_docs_ check should pass after merging https://github.com/2600hz/kazoo-crossbar/pull/304

**kazoo_endpoint/kz_privacy.erl**:
Currently, when _privacy method=kazoo_ and _privacy_ is enforced, both caller_id name and number are being anonymized correctly but the From-URI header is still including the caller_id’s number and realm there.

These changes take care of hiding information from From-URI field as well when required.

**kazoo_sip/kzsip_uri.erl**:
Update `kzsip_uri:parse/1` function in order to allow it to parse complex URIs, and opens the door to have different encoding configurations based on the URI type being encoded. So far, it only knows how to encode Request-URIs and From-URIs, but this list can be easily increased later as needed.

**kazoo_stdlib/kz_binary.erl**:
Adds 2 new functions to its API:
- `to_proplist/3`: Builds a proplist out of a binary.
- `matches/2`: Whether or not a regex matches the given subject.


kazoo_endpoint/src/kz_privacy.erl
```diff
@@ -226,6 +259,99 @@ hide_number(JObj, _Method) ->
...
+%% or any combination.
+%% @end
+%%------------------------------------------------------------------------------
+-spec hide_from_uri(kz_json:object(), kz_term:ne_binary()) -> kz_term:ne_binary().
+hide_from_uri(JObj, FromURI) ->
+    hide_from_uri(JObj, FromURI, [{'should_hide_realm', default_should_hide_realm()}]).
+
+-spec hide_from_uri(kz_json:object(), kz_term:ne_binary(), kz_json:proplist()) -> kz_term:ne_binary().
+hide_from_uri(JObj, FromURI, Options) ->
+    Regex = <<"(\")?([^\"<]*)(\"?\s+<sip:|\"\s+|^)([^@]*)@([^:;>]*)(.*)$">>,
```
lazedo: why are adding new ways of parsing a `uri` ? why not use kazoo_sip/kzsip_uri ?

k-anderson: The explanation was in the comment above the regex... but basically kzsip_uri can't handle all the use cases (also examples in the comments above the regex)

lazedo: can you provide and example that kzsip_uri can't handle ?

harenson: It does not handle _tag_ or anything else after the _port_ number:
```
kzsip_uri:parse(<<"number@realm.test:7000;tag=example">>).                                                  
** exception error: bad argument                                                                                                                 
     in function  binary_to_float/1                                                                                                              
        called as binary_to_float(<<"7000;tag=example">>)                                                                                        
     in call from kz_term:to_integer/2 (src/kz_term.erl, line 303)                                                                               
     in call from kzsip_uri:parse_host/2 (src/kzsip_uri.erl, line 90)
```

If name is also provided, both _name_ and _number_ are being taken as the _user_:
```
kzsip_uri:parse(<<""name" number@realm.test:7000">>).                                                       
* 1: syntax error before: name                                                                                                                   
(kazoo_apps@kazoo5x.harrylabs.com)3> kzsip_uri:parse(<<"\"name\" number@realm.test:7000">>).                                                     
{sip_uri,sip,<<"\"name\" number">>,<<"realm.test">>,7000}
```

lazedo: so maybe the change should be on kzsip_uri:parse

kazoo_endpoint/test/kz_privacy_tests.erl
```diff
@@ -0,0 +1,125 @@
...
+    kz_json:set_value(<<"Privacy-Hide-Name">>, ShouldHide, JObj).
+
+%% `Method=none' is only expected to be set by stepswitch_bridge when call is to an emergency number.
+%% Method: <<"sip">> | <<"none">> | <<"kazoo">>
+-spec set_method(kz_json:object(), kz_term:ne_binary()) -> kz_json:object().
+set_method(JObj, Method) ->
+    kz_json:set_value(<<"Privacy-Method">>, Method, JObj).
+
+-spec binary_replace(kz_term:ne_binary(), kz_term:proplist()) -> kz_term:ne_binary().
+binary_replace(Subject, []) ->
```
jamesaimonetti: why not a fold?
```erlang
binary_replace(Subject, Replacements) ->
    lists:foldl(fun({P, R}, S) -> binary:replace(S, P, R) end, Subject, Replacements).
```
Any particular reason to implement the fold directly vs lists:foldl/3?

kazoo_endpoint/test/kz_privacy_tests.erl
```diff
@@ -0,0 +1,125 @@
...
+    Routines = [{fun set_should_hide_number/2, 'false'}
+               ,{fun set_should_hide_name/2, 'false'}
+               ,{fun set_method/2, <<"sip">>}
+               ],
+    apply_kazoo_privacy(JObj, Routines).
+
+-type routine_value() :: boolean() | kz_term:ne_binary().
+-spec apply_kazoo_privacy(kz_json:object(), [{fun((kz_json:object(), routine_value()) -> kz_json:object()), routine_value()}]) -> kz_json:object().
+apply_kazoo_privacy(JObj, Routines) ->
+    lists:foldl(fun({F, Value}, Acc) -> F(Acc, Value) end, JObj, Routines).
```
jamesaimonetti: `kz_doc:setters(JObj, Routines)` has this ability already, no need to reimplement

kazoo_endpoint/src/kz_privacy.erl
jamesaimonetti: Agreed, other parts of the codebase could benefit from kzsip_uri being improved to parse these.

lazedo:
looks like the record needs two sets of params to support  `<sip:line1@192.0.2.2;transport=tcp>;reg-id=1;+sip.instance="<urn:uuid:00000000-0000-1000-8000-000A95A0E128>"`,

harenson:
@lazedo Would _transport_ be 1 of the 2 fields you mention? If so, what is the other one?

lazedo:
**any** param @harenson `<sip:line1@192.0.2.2;transport=tcp;nat=yes,from-offnet=true>;reg-id=1;+sip.instance="<urn:uuid:00000000-0000-1000-8000-000A95A0E128>`

kazoo_endpoint/test/kz_privacy_tests.erl
harenson: Not particular reason here, it was just what came to mind at that moment 😄

kazoo_endpoint/test/kz_privacy_tests.erl
harenson: Thank you for the heads up!

harenson:
@lazedo done 👍🏾

kazoo_sip/src/kzsip_uri.erl
```diff
@@ -111,23 +189,182 @@ scheme(#sip_uri{scheme=S}) -> S.
...
+    set_extra_param(Param, Value, Sip).
+
+-spec set_extra_param(uri_param_key(), uri_param_value(), sip_uri()) -> sip_uri().
+set_extra_param(Param, Value, #sip_uri{}=Sip) ->
+    set_extra_params(Sip, props:set_value(Param, Value, extra_params(Sip))).
+
+
+
+%% Helpers
+-spec build_sip_uri(kz_term:proplist()) -> #sip_uri{}.
```
jamesaimonetti: Use the `sip_uri()` type instead of the record

kazoo_sip/src/kzsip_uri.erl
```diff
@@ -111,23 +189,182 @@ scheme(#sip_uri{scheme=S}) -> S.
...
+build_sip_uri(PropsURI) ->
+    build_sip_uri(PropsURI, #sip_uri{}).
+
+-spec build_sip_uri(kz_term:proplist(), sip_uri()) -> #sip_uri{}.
```
jamesaimonetti: Fix second one here

kazoo_stdlib/src/kz_binary.erl
```diff
@@ -287,3 +321,10 @@ to_utf8(<<Value/binary>>) ->
             %% it must be a file's content
             Value
     end.
+
+-spec matches(kz_term:ne_binary(), kz_term:ne_binary()) -> kz_term:ne_binary().
```
jamesaimonetti: returns a `boolean()`

PR #442, opened 2022-06-08
Adding changes to send a conference email invite from branding app

kdamas93:

PR #452, opened 2022-06-29
Adding default headers

kdamas93:


icehess (approved):

icehess:
master: https://github.com/2600hz/kazoo-core/pull/442

PR #455, opened 2022-06-29
[Master] KZOO-264 : Add TCP and UDP toggle for DNS Query

sanjeevie6353:
Added changes to allow TCP query for DNS lookup. Default is TCP query.

Linked changes available on both kazoo-crossbar & kazoo-ecallmgr under branch name KZOO-264

https://github.com/2600hz/kazoo-crossbar/pull/308
https://github.com/2600hz/kazoo-ecallmgr/pull/171


lazedo (changes requested):
you're adding dependency for kapps_config in this library, is there a reason for this ? applications using network utils will have kapps_config dependency which is not intended on some application that don't use kazoo_apps/kazoo_appex. you can query/setup the option on appex and pass it to network utils

kazoo_appex/src/appex_bluepak.erl
```diff
@@ -300,10 +301,17 @@ write_path(Filename, #load_app{app_dir=AppDir
 api_url() ->
     case appex_client_config:api_url() of
         'undefined' ->
+            Options = 
+                case kapps_config:is_true(?CONFIG_CAT, <<"use_tcp">>, 'false') of
+                    'true' -> kz_network_utils:default_options();
+                    'false' -> props:set_value('usevc', 'false', kz_network_utils:default_options())
+                end,
```
lazedo: move this to a fun

kazoo_appex/src/appex_bluepak.erl
```diff
@@ -300,10 +301,17 @@ write_path(Filename, #load_app{app_dir=AppDir
...
+
             try kz_network_utils:resolve(
                   kz_term:to_binary(
                     kz_network_utils:get_hostname()
                    )
+                  , Options
```
lazedo: invoke the fun here

kazoo_stdlib/src/kz_network_utils.erl
```diff
@@ -580,7 +580,7 @@ new_options() -> [].
 
 -spec default_options() -> options().
 default_options() -> ?LOOKUP_OPTIONS.
-
+
```
lazedo: remove this change (spaces added)

kazoo_stdlib/src/kz_network_utils.erl
```diff
@@ -647,4 +647,4 @@ cleanse_mac_address(MACAddress) ->
             end;
         'false' ->
             'undefined'
-    end.
```
lazedo: remove this change too

kazoo_appex/src/appex_bluepak.erl
sanjeevie6353: Moved to a function.
image

kazoo_appex/src/appex_bluepak.erl
sanjeevie6353: Invoked it using the above function

kazoo_stdlib/src/kz_network_utils.erl
sanjeevie6353: removed the change

kazoo_stdlib/src/kz_network_utils.erl
sanjeevie6353: removed this change.

icehess:
@sanjeevie6353 Isn't it better to move all these `use_tcp` configs from individual places to `kz_network_utils` as a global settings?

I think having separate config for this is excessive and cumbersome, generally if you need this option it needed by most of other places in Kazoo too, by having a global settings it is easier as admin of system to set it instead of forgetting to set this option in appex (which virtually all people will forget) but having setting in cb_whitelable.

icehess:
Or let ecallmgr, appex_client, cb_whitelabel fetch their ow `use_tcp` option, but remove their default value
and in `kz_network_util` if `usevc` is `undefined` then it fetches the global settings of `use_tcp`.

jamesaimonetti:
@icehess kz_network_utils, being in kazoo_stdlib, should not use kapps_config. Each app that uses these should override the default 'true' if they choose.

icehess:
@jamesaimonetti fair enough.

PR #462, opened 2022-07-19
[master] KCRO-146: generate and merge keys into bookmark

icehess:
`kz_couch_view` have to properly generate and merge bookmark, otherwise it would missing setting `endkey`  and view may return faulty result.

If `kz_couch_view` doesn't use view module or options has key/range keys then uses those options as-is, otherwise generate keys using selector as before and merge bookmark into it.

Also we do a sanity checks:

1. Check if key (or keys) and range keys are define together. This is an invalid option and CouchDB will return error
2. Check if option is key/keys, in this case we have to turn off batching because we don't have any `endkey` and batching depends on setting `startkey` on each`next_batch`. This to avoid re-introducing the same problem that this PR is fixing
3. Converting or improving the all codes that are using key/keys, and either trun off pagination/batching or convert them to use same start/end keys (mostly codes in crossbar)


icehess:
Required PR: https://github.com/2600hz/kazoo-crossbar/pull/319

Fix couchbeam types: https://github.com/2600hz/kazoo5/pull/81

icehess:
Backported PRs 5.1:

* core-5.1: https://github.com/2600hz/kazoo-core/pull/473
* crossbar-5.1: https://github.com/2600hz/kazoo-crossbar/pull/328

Backported PRs 5.0:

* core-5.0: https://github.com/2600hz/kazoo-core/pull/474
* crossbar-5.0: https://github.com/2600hz/kazoo-crossbar/pull/329

PR #468, opened 2022-07-28
[Master] KTAS-2 : Update 'modb' task to allow optional archiving

sanjeevie6353:
Added config option to allow deleting the 'modb' without archiving. default is false.

Changes on kazoo-core, kazoo-tasks and kazoo-crossbar.


kazoo_modb/src/kazoo_modb.erl
```diff
@@ -570,12 +570,21 @@ migrate_routines([R | Rs], Acc) ->
...
+
+-spec maybe_archive_modb(kz_term:ne_binary(), boolean()) -> 'ok'.
+maybe_archive_modb(AccountMODb, ArchiveBeforeDel) ->
     {Year, Month, _} = erlang:date(),
     case should_archive(AccountMODb, Year, Month) of
         'true' ->
-            lager:info("account modb ~s needs archiving", [AccountMODb]),
-            'ok' = kz_datamgr:db_archive(AccountMODb),
-            lager:info("account modb ~s archived, removing the db", [AccountMODb]),
+            case ArchiveBeforeDel of
```
jamesaimonetti: I think you can just use boolean logic in one case statement:

```erlang
case should_archive(AccountMODb, Year, Month)
   andalso ArchiveBeforeDel
of
    'true' -> kz_datamgr:db_archive();
    'false' -> lager:info("not archiving first")
end
```

jamesaimonetti (changes requested):

kazoo_modb/src/kazoo_modb.erl
jamesaimonetti: | should_archive/3 | ArchiveBeforeDel | kz_datamgr:db_archive/0 |
|------------------------|-------------------------|-----------------------------------|
| true | true | true |
| true | false | false |
| false | true | false |
|false | false | false |

sanjeevie6353: > I think you can just use boolean logic in one case statement:
> 
> ```erlang
> case should_archive(AccountMODb, Year, Month)
>    andalso ArchiveBeforeDel
> of
>     'true' -> kz_datamgr:db_archive();
>     'false' -> lager:info("not archiving first")
> end
> ```

I have a small concern on this.

- If the existing function should_archive/3 resolves to false, there will be no modb deletion
- If should_archive/3 resolves to true, the function needs to execute the db_delete (irrespective of the value of <<"archive_modb_before_delete">>) parameter.

So we might have to keep the current implementation.

PR #469, opened 2022-08-02
[master] AXS-41: use custom ensure_all_started to avoid race condition on app start

icehess:
Erlang `application:ensure_all_started/1` will do a `application:stop/1` when
the app or one its dependecies errors out during start. The error could be
due to the fact the a dependecy is not locally available.

In the case of missing dependecy kapps_controller will try external source again
to load the dep.

This stop, and try load external along with any cleaning up by external source
will have race condition and a loop of wiping and re-fetching the same set of apps.

Here we use a custom ensure_all_started lifted from the `application` module to
specifically handle missing deps and give external source a chance to load
everything up and try the sart the app. We only stop all started apps on other
errors.


icehess:
5.1 PR: https://github.com/2600hz/kazoo-core/pull/470

PR #471, opened 2022-08-02
[master] spliting appex http client from controller and cleanup

icehess:


icehess:
5.1 PR: https://github.com/2600hz/kazoo-core/pull/472

PR #476, opened 2022-08-05
KZOO-284: only send service_added notification when the change is result of direct updates by user

icehess:
For whatever reason services quantity may not be properly saved. So when a user triggers a billable API request (like creating a device), crossbar will run services which will include the previously uncounted quantity. The a the notification will alarm user to a change not related to their actual recent API request.

This tries to reduce and avoid this situation. During item annotation we check if there is any update for this category/item (previously noted by `kz_services_quantities:calculate_updates/3` for Current and Proposed JObj). If there is an update we mark this change due to update to their services. Otherwise we consider the change as a result of sync/reconcile (or updates to the plans).

Then during sending notification, this checks the changes are result of an update or not. If there is no update at all nothing will be send. If there are updates, there may a chance that some items are still re-counted and changed due to reconciliation. For this reason all items check and if they are not explicitly change as a result of updates their changed are nullified and teletype won;t mark them as changed in the email and this hopefully reduces confusion to the users.


icehess:
PRs:
- master: https://github.com/2600hz/kazoo-core/pull/476
- 5.4: https://github.com/2600hz/kazoo-core/pull/962

PR #498, opened 2022-09-15
KCRO-150: Enhance DELETE /users/{USER_ID} API

harenson:
Dependency for https://github.com/2600hz/kazoo-crossbar/pull/344
Tests https://github.com/2600hz/kazoo-properly/pull/93

With this change, when deleting a user, either "all" or a list of crossbar's `object_types` may be included in the request (either in the request's body or query_string) for the API to take of deleting those along the user. There is also the possibility to include
"phone_numbers" in the objects list, which will instruct the users API to also release all the numbers associated to every callflow owned by the user.


kazoo_stdlib/src/props.erl
```diff
@@ -300,6 +304,15 @@ take_value(Key, Props, Default) ->
         Value -> {Value, delete(Key, Props)}
     end.
 
+%%------------------------------------------------------------------------------
+%% @doc Splits given proplist based on filter function (Fun).
+%% @end
+%%------------------------------------------------------------------------------
+-spec split(filter_fun(), kz_term:proplist()) -> {kz_term:proplist(), kz_term:proplist()}.
+split(Fun, Props) ->
```
jamesaimonetti: This is an interesting decision. Compare 10 runs on a list of 10,000 elements:
```
split() ->
    props:split(fun is_true/1, ?L).

split_p() ->
    lists:splitwith(fun is_true/1, ?L).

props_tests:split in 1.545273s
props_tests:split_p in 0.000000s
```

jamesaimonetti: Correction, use `lists:partition/2` but same timings apply:
```
props_tests:split in 1.545738s
props_tests:split_p in 0.002573s
```

jamesaimonetti: 2nd point, tests for this functionality should be included in props_tests. I added a commit with a couple

kazoo_couch/priv/couchdb/views/account-crossbar_listings.json
```diff
@@ -24,6 +24,71 @@
                 "}"
             ]
         },
+        "by_ownerid": {
```
icehess: Why creating a new view? What functionality `by_ownerid_type` is not providing which makes you create new view?

If you just need to search for all docs owned by owner_id simply use `[{startkey, [OwnerId]}, {endkey, [OwnerId, kz_term:high_unicode_value()]}]`.

icehess (changes requested):

kazoo_stdlib/src/props.erl
harenson: Ok, thank you for the heads up.

kazoo_couch/priv/couchdb/views/account-crossbar_listings.json
harenson: Good to know, thanks @icehess, I just changed it.

PR #508, opened 2022-10-10
[5.1] SHIP-42: migrate settings from old default url to new branch scheme

swysor:


jamesaimonetti:
Assuming 5.1 isn't needed, reopen if not

PR #515, opened 2022-10-26
[5.1] KZOO-314: only rename app if the app is already installed in the account

icehess:
Rename maintenance command should check if the app is NOT installed in account.

Rename blindly without this check will activate the app for all accounts. Also since we don't properly set the required fields in apps_store, any future changes in apps_store doc will cause crashes during service run.

Also adding a maintenance command to fix the problem introduced by rename before.


icehess:
master pr: https://github.com/2600hz/kazoo-core/pull/514

PR #516, opened 2022-10-26
KCRO-160: update accessors for add options to music_on_hold

icehess:


icehess:
crossbar pr: https://github.com/2600hz/kazoo-crossbar/pull/367

icehess:
closing in favor of new pr that Luis is going to make

PR #524, opened 2022-11-09
kcro-148-CommlandSelectorApi-master

kdamas93:
Related changes for command selector API support, more detail can be found in this reference
https://github.com/2600hz/kazoo-crossbar/pull/374


kazoo_documents/src/kzd_application_blocklist.erl
```diff
@@ -0,0 +1,35 @@
+%%%-------------------------------------------------------------------
+%%% @author kevin
+%%% @copyright (C) 2022, <COMPANY>
+%%% @doc Module for creating application_block list
+%%% @end
+%%% Created : 01. Nov 2022 13:06
+%%%-------------------------------------------------------------------
+-module(kzd_application_blocklist).
+-author("kevin").
```
jamesaimonetti: uneeded

kazoo_documents/src/kzd_application_blocklist.erl
```diff
@@ -0,0 +1,35 @@
+%%%-------------------------------------------------------------------
+%%% @author kevin
```
jamesaimonetti: Please find a module header comment block and re-use here

kazoo_documents/src/kzd_application_blocklist.erl
```diff
@@ -0,0 +1,35 @@
...
+
+-define(ID, <<"application_blocklist">>).
+
+%%------------------------------------------------------------------------------
+%% @doc Create a new application_blocklist for the requested account
+%% @end
+%%------------------------------------------------------------------------------
+-spec new(kz_term:ne_binary(), kz_term:ne_binary()) -> kz_json:object().
+new(<<Account/binary>>, ApplicationType) ->
+    Routines = [fun(JObj) -> kz_doc:set_id(JObj, ?ID) end
```
jamesaimonetti: See about using kz_json:exec_first/2 instead, then you can write it like:

```erlang
    Routines = [{fun kz_doc:set_id/2, ?ID}
               ,{fun kz_doc:set_account_id/2, kzs_util:format_account_id(Account)}
               ,{fun kz_doc:set_account_db/2, kzs_util:format_account_db(Account)}
               ,fun(JObj) ->
                        kz_json:set_value(ApplicationType, [], JObj)
                end
               ],
    kz_json:exec_first(Routines, kz_json:new()).
```

As a general rule, we try to use named functions as arguments instead of contructing anonymous funs (when possible). There's a performance penalty to building the fun then executing it vs passing a named fun and executing it.

kazoo_documents/src/kzd_application_entitlement.erl
```diff
@@ -0,0 +1,32 @@
...
+
+-define(ID, <<"application_entitlement">>).
+
+%%------------------------------------------------------------------------------
+%% @doc
+%% @end
+%%------------------------------------------------------------------------------
+-spec new(kz_term:ne_binary()) -> kz_json:object().
+new(<<Account/binary>>) ->
+    Routines = [fun(JObj) -> kz_doc:set_id(JObj, ?ID) end
```
jamesaimonetti: Same idea here

jamesaimonetti (changes requested):

kazoo_documents/src/kzd_application_blocklist.erl
kdamas93: Done.

kazoo_documents/src/kzd_application_blocklist.erl
kdamas93: Done ! Thanks!

kazoo_documents/src/kzd_application_entitlement.erl
kdamas93: Done ! Thanks!

kazoo_documents/src/kzd_application_blocklist.erl
```diff
@@ -0,0 +1,30 @@
...
+%% @doc Create a new application_blocklist for the requested account
+%% @end
+%%------------------------------------------------------------------------------
+-spec new(kz_term:ne_binary(), kz_term:ne_binary()) -> kz_json:object().
+new(<<Account/binary>>, ApplicationType) ->
+    Routines = [{fun kz_doc:set_id/2, ?ID}
+               ,{fun kz_doc:set_account_id/2, kzs_util:format_account_id(Account)}
+               ,{fun kz_doc:set_account_db/2, kzs_util:format_account_db(Account)}
+               ,fun(JObj) ->
+                        kz_json:set_value(ApplicationType, [], JObj)
```
icehess: add an accessor and setter function for `application_type` here, and use it here instead of anonymous function.

icehess (changes requested):

kazoo_documents/src/kzd_application_blocklist.erl
kdamas93: @icehess change made, thanks!

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -64,11 +66,21 @@ init_apps(AppsPath, ApiUrl) ->
...
+init_application(AppPath) ->
+    init_application(AppPath, 'undefined').
+
 -spec init_app(file:filename_all(), kz_term:api_binary()) -> 'ok'.
 init_app(AppPath, ApiUrl) ->
     Keys = [<<"api_url">>],
     init_app(AppPath, ApiUrl, Keys).
 
+-spec init_application(file:filename_all(), kz_term:api_binary()) -> 'ok'.
+init_application(AppPath, AppUrl) ->
```
icehess: What is `AppUrl`?

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -245,11 +257,87 @@ maybe_update_app(AppPath, ApiUrl, Keys) ->
...
+            ?SUP_LOG_INFO("  failed to find metadata in ~s: ~p~n", [AppPath, _E])
+    end.
+
 -spec maybe_set_api_url(kz_term:api_binary(), kz_json:object()) -> kz_json:object().
 maybe_set_api_url('undefined', MetaData) ->
     kz_json:delete_key(<<"api_url">>, MetaData);
-maybe_set_api_url(ApiUrl, MetaData) ->
-    kz_json:set_value(<<"api_url">>, ApiUrl, MetaData).
+maybe_set_api_url(AppUrl, MetaData) ->
+    kz_json:set_value(<<"api_url">>, AppUrl, MetaData).
```
icehess: This is really crossbar Api url, what is this `AppUrl`?

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -245,11 +257,87 @@ maybe_update_app(AppPath, ApiUrl, Keys) ->
...
+
+-spec find_metadata_application(file:filename_all()) ->
+          {'ok', kz_json:object()} |
+          {'invalid_data', kz_term:proplist()}.
+find_metadata_application(AppPath) ->
+    {'ok', Bin} = file:read_file(filename:join([AppPath, <<"metadata">>, <<"app.json">>])),
+    JSON = kz_json:decode(Bin),
+    case kz_json_schema:validate(<<"application">>, kz_doc:public_fields(JSON)) of
```
icehess: I can find the `application` schema, is this new? How it is different from old `app`?

Also there is collision between application schema and application schema from kazoo-cpaas (which is totally different thing)

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -245,11 +257,87 @@ maybe_update_app(AppPath, ApiUrl, Keys) ->
...
+        {'ok', _} -> {'ok', JSON};
```
icehess:
```suggestion
        {'ok', _} -> {'ok', kzd_app:set_app_hash(JSON, kz_binary:rand_hex(16))};
```

icehess (changes requested):
I still don't understand why we need another function to init app.
We already have a proven working set of functions, like `init_apps`, `init_app`, `refresh_app`, `refresh_apps` and their rename functions. 

this only adds init_app.

What concerns me here that implementing this feature this way, it is not clear in what circumstances we should use init_app or init_applications, and this is highly confusing both to other developers and most importantly to ops and system_admins. Also give that there is no documentation, no spec and no comments.

What I would suggest to incorporate `application_type` to `metadata.json` of apps,
Commland apps, naturally add this field as they add the metadata.json. MUI apps will add or updates in the future.

By default for now, if application_type is missing we assume this monster-ui-app to make current non-updated mui metadata compatible and put less pressure in developers and system_admin about migration.

This there is absolutely no need to add these new functions, you just update functions like `find_metadata_application` and `maybe_update_app` to incorporate the application_type logic.

Also this way we don't need so much different copy of the same thing, like view or other functions/modules/schemas.

Basically what I'm asking is just add application_type to the existing `app` schema and update the logic to check and honor application_type in different places like here.

Feel free to reach out to me if you have questions.

kazoo_documents/src/kazoo_documents_maintenance.erl
```diff
@@ -280,6 +344,20 @@ create_app(AppPath, MetaData, MasterAccountDb) ->
                          )
     end.
 
+-spec create_application(file:filename_all(), kz_json:object(), kz_term:ne_binary()) -> 'ok'.
+create_application(AppPath, MetaData, MasterAccountDb) ->
+    Doc0 = kz_doc:update_pvt_parameters(MetaData, MasterAccountDb, [{'type', <<"application">>}]),
+    Doc = kz_json:delete_key(<<"source_url">>, Doc0),
```
icehess: Why deleting this? Removing this will break kazoo_appex.

icehess (changes requested):

kazoo_documents/src/kazoo_documents_maintenance.erl
kdamas93: @icehess I will squash all the commits to work based on a single commit and after that I will work/respond on your observations, thanks for the explanation

kazoo_documents/src/kzd_application_blocklist.erl
```diff
@@ -0,0 +1,37 @@
...
+%%------------------------------------------------------------------------------
+%% @doc Create a new application_blocklist for the requested account
+%% @end
+%%------------------------------------------------------------------------------
+-spec new(kz_term:ne_binary(), kz_term:ne_binary()) -> kz_json:object().
+new(<<Account/binary>>, ApplicationType) ->
+    Routines = [{fun kz_doc:set_id/2, ?ID}
+               ,{fun kz_doc:set_account_id/2, kzs_util:format_account_id(Account)}
+               ,{fun kz_doc:set_account_db/2, kzs_util:format_account_db(Account)}
+               ,{fun set_application_type/2, ApplicationType}
```
icehess: hmm, also do we want to set `doc_type` here?

kazoo_documents/src/kzd_application_entitlement.erl
```diff
@@ -0,0 +1,27 @@
...
+
+%%------------------------------------------------------------------------------
+%% @doc Create a new application_entitlement for the requested account
+%% @end
+%%------------------------------------------------------------------------------
+-spec new(kz_term:ne_binary()) -> kz_json:object().
+new(<<Account/binary>>) ->
+    Routines = [{fun kz_doc:set_id/2, ?ID}
+               ,{fun kz_doc:set_account_id/2, kzs_util:format_account_id(Account)}
+               ,{fun kz_doc:set_account_db/2, kzs_util:format_account_db(Account)}
```
icehess: hmm, also do we want to set `doc_type` here?

icehess (changes requested):

kazoo_documents/src/kazoo_documents_maintenance.erl
kdamas93: should be `ApplicationUrl`

kazoo_documents/src/kazoo_documents_maintenance.erl
kdamas93: I just uploaded it, it's based on the app schema, it's basically a copy to isolate the new functionality from the existing one

kazoo_documents/src/kazoo_documents_maintenance.erl
kdamas93: I have not deleted it, when it does not appear it was because it was located in the wrong place, I had created it in crossbar but it is already in core

kazoo_documents/src/kzd_application_entitlement.erl
kdamas93: I don't quite understand, how should the change be or is it unnecessary to use a macro?

kdamas93:
> I still don't understand why we need another function to init app. We already have a proven working set of functions, like `init_apps`, `init_app`, `refresh_app`, `refresh_apps` and their rename functions.
> 
> this only adds init_app.
> 
> What concerns me here that implementing this feature this way, it is not clear in what circumstances we should use init_app or init_applications, and this is highly confusing both to other developers and most importantly to ops and system_admins. Also give that there is no documentation, no spec and no comments.
> 
> What I would suggest to incorporate `application_type` to `metadata.json` of apps, Commland apps, naturally add this field as they add the metadata.json. MUI apps will add or updates in the future.
> 
> By default for now, if application_type is missing we assume this monster-ui-app to make current non-updated mui metadata compatible and put less pressure in developers and system_admin about migration.
> 
> This there is absolutely no need to add these new functions, you just update functions like `find_metadata_application` and `maybe_update_app` to incorporate the application_type logic.
> 
> Also this way we don't need so much different copy of the same thing, like view or other functions/modules/schemas.
> 
> Basically what I'm asking is just add application_type to the existing `app` schema and update the logic to check and honor application_type in different places like here.
> 
> Feel free to reach out to me if you have questions.

This design was discussed in its initial version with Karl.

The purpose is to create a new `application` schema based on the current app schema.

I just added` init-application` to complete the application creation flow through the new `application` schema.
I understand that this is the idea to be able to differentiate or that there would not be problems of conflicts about the already existing functionality.
I understand that I would need to add the functions `init_applications`, `refresh_aplication`, `refresh_applications`

I can add the necessary/missing documentation but I would like to confirm if we will keep this architecture or choose to always handle it in the application schema with a new section in the metadata.

How would we reach an agreement or is the version proposed by you the one that we should follow?

I appreciate your time and level of detail, I am waiting for your comments and in parallel I will send you this answer in slack

kazoo_documents/src/kzd_application_blocklist.erl
icehess: Also please write JSON schema for this.

kazoo_documents/src/kzd_application_entitlement.erl
icehess: ALso please write JSON schema.

icehess (changes requested):

kazoo_documents/src/kazoo_ui_apps_maintenance.erl
```diff
@@ -0,0 +1,792 @@
...
+            rename_app(kz_doc:id(AppDoc), PrevName, kzd_app:name(MetaData));
+        {'error', Error} ->
+            ?SUP_LOG_INFO(" failed to find app ~s: ~p", [PrevName, Error])
+    end.
+
+-spec rename_app(kz_term:ne_binary(), kz_term:ne_binary(), kz_term:ne_binary()) -> 'ok'.
+rename_app(AppId, PrevName, NewName) ->
+    {'ok', MA} = kapps_util:get_master_account_id(),
+    Accounts = kapps_util:account_descendants(MA),
+    case rename_app('legacy', AppId, PrevName, NewName, [MA | Accounts]) of
```
jamesaimonetti: Either this first arg `legacy` is wrong or the type in `rename_app/5` is wrong (tuple or ne_binary)

jamesaimonetti: A question here too: in `rename_app/5` `AppType` is never used for anything, so why pass it around?

kazoo_documents/src/kazoo_ui_apps_maintenance.erl
```diff
@@ -0,0 +1,792 @@
...
+        {'ok', AppJObj} ->
+            ?SUP_LOG_INFO(" app ~s already loaded in system~n", [AppName]),
+            maybe_update_app({'legacy', <<"monster-ui">>}, AppPath, MetaData, Keys, MasterAccountDb, AppJObj);
+        {'error', 'not_found'} -> create_app({'legacy', <<"monster-ui">>}, AppPath, MetaData, MasterAccountDb);
+        {'error', _E} -> ?SUP_LOG_INFO(" failed to find app ~s: ~p", [AppName, _E])
+    end;
+maybe_create_app(AppType, AppPath, MetaData, Keys, MasterAccountDb) ->
+    AppName = kzd_app:name(MetaData),
+    lager:warning("~p ~p",[AppName, MasterAccountDb]),
+    case kzd_application:find_by_name(MasterAccountDb, AppName, AppType) of
```
jamesaimonetti: You may want to put `<<AppType/binary>>` in the function clause to let dialyzer know, otherwise it complains that this function call won't work.

kazoo_documents/src/kazoo_ui_apps_maintenance.erl
```diff
@@ -0,0 +1,792 @@
...
+update_app(AppId, Path, Value) ->
+    case lists:last(Path) of
+        <<"_id">> -> 'ok';
+        <<"pvt_", _/binary>> -> 'ok';
+        ?NE_BINARY ->
+            {'ok', MA} = kapps_util:get_master_account_db(),
+            Updates = [{Path, Value}],
+            UpdateOptions = [{'update', Updates}],
+            case kz_datamgr:update_doc(MA, AppId, UpdateOptions) of
+                {'ok', _} -> app('legacy',AppId);
```
jamesaimonetti: I don't think the atom here is valid, right? `app/2` expects the tuple or ne_binary()

kazoo_documents/src/kazoo_ui_apps_maintenance.erl
```diff
@@ -0,0 +1,792 @@
...
+set_app_screenshots(_AppType, AppId, PathToScreenshotsFolder) ->
+    {'ok', MA} = kapps_util:get_master_account_db(),
+    ?SUP_LOG_INFO("processing...\n"),
+    SShots = [{filename:basename(SShot), SShot}
+              || SShot <- filelib:wildcard(kz_term:to_list(PathToScreenshotsFolder) ++ "/*.png")
+             ],
+    update_screenshots(AppId, MA, SShots).
+
+-spec fix_apps_store_SUPP_212(tuple() | kz_term:ne_binary(), kz_term:ne_binary() | kz_term:ne_binaries()) -> 'ok'.
+fix_apps_store_SUPP_212(_AppType, AppId) ->
```
jamesaimonetti: Does this function call itself infinitely? The function is arity-2 and calls the same named function, arity 2, with diff args.

What tests do you have calling this function to show it works?

kazoo_documents/src/kzd_application.erl
```diff
@@ -0,0 +1,302 @@
...
+-spec maybe_get_app_name(list(), kz_term:ne_binary(), kz_term:ne_binary()) -> {boolean(), kz_term:ne_binary()}.
+maybe_get_app_name([App | Apps], AppName, AppType) ->
+    Name = kz_json:get_value(<<"name">>, App),
+    Type = kz_json:get_value(<<"type">>, App),
+    StringAppType = list_to_binary(AppType),
+    case Name =:= AppName
+        andalso Type =:= StringAppType
+    of
+        'true' ->
+            {'ok', App};
```
jamesaimonetti: why return the ok tuple when the spec says boolean() and the calling code above pattern matches for boolean?

I ask again, what tests use this code?

jamesaimonetti (changes requested):
See crossbar review as well

kazoo_documents/src/kazoo_ui_apps_maintenance.erl
kdamas93: It is a function that has been inherited from **kazoo_documents_maintenance**, it was taken since the development of this module is based on the **kazoo_documents_maintenance**

icehess:
Required PRs:

- https://github.com/2600hz/kazoo-crossbar/pull/374
- https://github.com/2600hz/kazoo-core/pull/524
- https://github.com/2600hz/kazoo-properly/pull/156
- https://github.com/2600hz/kazoo-ast/pull/42

PR #525, opened 2022-11-10
Ignore DNS lookups when testing

jamesaimonetti:


jamesaimonetti:
Look at https://github.com/2600hz/kazoo-properly/pull/104 too

kazoo_schemas/src/kz_json_schema_extensions.erl
```diff
@@ -360,6 +360,12 @@ is_valid_port(Port, BlockedPorts) ->
             'false'
     end.
 
+-ifdef(TEST).
+%% when running local tests don't check DNS for hostnames
+is_valid_dns(_Host) -> 'true'.
```
harenson, line 365: I think you need to add a spec for this function too, otherwise `compile-test[-direct]` rule(s) will complain.

jamesaimonetti: not exported so not required to have a spec tho

harenson: Oh, you are correct 👍🏾

jamesaimonetti:
Fix for multiple schema errors being lost: https://github.com/2600hz/kazoo-crossbar/pull/376

PR #528, opened 2022-11-10
KZOO-321: enable using AWS EC2 instance profiles for S3 att handler auth

danielfinke:
- instance metadata credentials are only used if permitted on the node, for security reasons
  - if disallowed, att handler returns 403 Forbidden

kazoo-crossbar PR: https://github.com/2600hz/kazoo-crossbar/pull/379


kazoo_attachments/src/kz_att_s3.erl
```diff
@@ -22,22 +22,33 @@
...
+
 -define(AMAZON_S3_HOST, <<"s3.amazonaws.com">>).
 -define(AMAZON_S3_UPLOAD_TIMEOUT, ?MILLISECONDS_IN_MINUTE * 30).
 
 -define(AMAZON_S3_HTTP_DEFAULT_CLIENT, 'lhttpc').
 %% httpc seems broken in otp 21.2.5
 
+-define(HTTP_RESP_CODE_FORBIDDEN, 403).
+
+-type bpc() :: {string(), kz_term:api_ne_binary(), aws_config()}.
```
jamesaimonetti, line 38: What is a `bpc`? Can we either have a nicely named type or a comment explaining what this tuple is about?

PR #531, opened 2022-11-15
KCAL-45 Add aliases to the voicemail box

nkarunarathne:
Previously the voicemail box had only a voicemail box number for box identification. It was requested add new aliases so that the voicemail box can be looked up using either the voicemail number or aliases.

The new 'aliases' parameter is a list of strings that can be used to identify a voicemail box uniquely. Each voicemail box number and/or aliases are always unique in the account.

Also, The voicemail aliases can be used as transfer targets. The voicemail aliases can be used when calling to check voicemail where the voicemail box number is asked for.

The Code Level Change

Edited the existing listing_by_mailbox view so that it would check both voicemail box IDs and aliases when doing the lookups.
Check the uniqueness of both voicemail box IDs and aliases when receiving API requests.
The voicemail box lookup was using an integer box number, changed it to binary. 

Changes are in
crossbar, callflow, core


kazoo_documents/src/kzd_vmboxes.erl
```diff
@@ -100,6 +101,18 @@ after_forward_action(Doc, Default) ->
 set_after_forward_action(Doc, ForwardType) ->
     kz_json:set_value(<<"after_forward_action">>, ForwardType, Doc).
 
+-spec aliases(doc()) -> kz_term:api_ne_binary().
+aliases(Doc) ->
+    aliases(Doc, 'undefined').
+
+-spec aliases(doc(), Default) -> kz_term:ne_binary() | Default.
+aliases(Doc, Default) ->
+    kz_json:get_ne_binary_value([<<"aliases">>], Doc, Default).
```
icehess: Aliases is a list, no?
then use this:

```suggestion
    kz_json:get_ne_binaries_value([<<"aliases">>], Doc, Default).
```

icehess (changes requested):

icehess:
@nkarunarathne 

You also need to modify these crossbar views, search for vmbox cases and add aliases to mailbox if it needs:

* `core/kazoo_couch/priv/couchdb/views/account-crossbar_listings.json` in kazoo-core
* `applications/locations/priv/couchdb/views/account-locations.json` in kazoo-locations repo

kazoo_documents/src/kzd_vmboxes.erl
```diff
@@ -100,6 +101,18 @@ after_forward_action(Doc, Default) ->
 set_after_forward_action(Doc, ForwardType) ->
     kz_json:set_value(<<"after_forward_action">>, ForwardType, Doc).
 
+-spec aliases(doc()) -> kz_term:api_ne_binary().
+aliases(Doc) ->
+    aliases(Doc, 'undefined').
+
+-spec aliases(doc(), Default) -> kz_term:ne_binary() | Default.
```
icehess:
```suggestion
-spec aliases(doc(), Default) -> kz_term:ne_binaries() | Default.
```

kazoo_documents/src/kzd_vmboxes.erl
```diff
@@ -100,6 +101,18 @@ after_forward_action(Doc, Default) ->
 set_after_forward_action(Doc, ForwardType) ->
     kz_json:set_value(<<"after_forward_action">>, ForwardType, Doc).
 
+-spec aliases(doc()) -> kz_term:api_ne_binary().
+aliases(Doc) ->
+    aliases(Doc, 'undefined').
```
icehess: Let use empty list as default values. This reads better and also people know that is a list.

```suggestion
-spec aliases(doc()) -> kz_term:ne_binaries().
aliases(Doc) ->
    aliases(Doc, []).
```

icehess (changes requested):
some small change reqeusts.

kazoo_documents/src/kzd_vmboxes.erl
icehess: also please be aware that if you change the default value also change the default value in schema in crossbar PR.

kazoo_couch/priv/couchdb/views/account-crossbar_listings.json
```diff
@@ -643,9 +643,12 @@
                 "",
                 "      /* attributes/mailbox_number */",
                 "      /* vmboxes/listing_by_mailbox */",
-                "      additionalEmitList.push({",
-                "        'key': [doc.pvt_type, doc.mailbox]",
-                "      });",
+                "      var aliases = doc.aliases || [];",
+                "      for (var element in aliases.push(doc.mailbox)) {",
```
danielfinke: `Array.prototype.push` returns a new length value, not a new array, so this doesn't seem like it will work correctly. I don't see any `"vmbox"` entries in the view

danielfinke (changes requested):

kazoo_couch/priv/couchdb/views/account-crossbar_listings.json
icehess: good catch, he probably meant to write: 

```suggestion
                "      for (var element in aliases.concat(doc.mailbox)) {",
```

icehess (changes requested):

danielfinke (approved):

PR #534, opened 2022-11-15
KREG-3: Move softphone to teammate migration to kazoo_endpoint

harenson:
Dependency for https://github.com/2600hz/kazoo-registrar/pull/20

Allows to migrate a softphone device to a teammate device by setting type=teammate and also sets default teammate values (codecs and other settings) to device.


lazedo (changes requested):
create a `module` for this teams integration and move it elsewhere, not `kazoo_endpoint`. maybe kapps_maintenance. why is this needed anyway ?

harenson:
@lazedo This is a temporary functionality, because there are already several accounts using softphone devices as MS Teams devices, and the idea is to avoid all those devices having to be deleted and then created as MS Teams devices once this enforcement is deployed.

harenson:
Not needed anymore, migration process was moved back into registrar's PR.

PR #536, opened 2022-11-22
KZOO-323: add support for user-defined metadata on call recordings

danielfinke:
- add `gen_attachment:update_metadata/5` callback
  - no-op if unimplemented
- `kz_att_s3` uses `tags_key_path` to find object tags from metadata and
  updates S3 object
- fix some types in `kapps_account_config`

kazoo5 PR: https://github.com/2600hz/kazoo5/pull/87
kazoo-core PR 2: https://github.com/2600hz/kazoo-core/pull/550
kazoo-crossbar PR: https://github.com/2600hz/kazoo-crossbar/pull/386
kazoo-media-mgr PR: https://github.com/2600hz/kazoo-media-mgr/pull/29


kazoo_data/src/kzs_attachments.erl
```diff
@@ -269,6 +270,67 @@ publish_doc(Plan, DbName, DocId) ->
...
+%% @end
+%%------------------------------------------------------------------------------
+-spec update_metadata(map()
+                     ,gen_attachment:db_name()
+                     ,gen_attachment:doc_id()
+                     ,gen_attachment:att_name()
+                     ,kz_doc:doc()
+                     ) -> gen_attachment:update_metadata_response().
+update_metadata(Plan, DbName, DocId, AName, Doc) ->
+    do_on_existing_att(Plan
```
icehess: If I'm reading the code right, the logic here is simple, but the implementation make it complicated and spread over multiple functions. I had to scroll and find and read the function to understand what is going on.

Simpler way is a simple traditional case (imho):

* the bulk of logic is in `do_on_existing_att`, so move its content here under this function 
* if no handler return ok (straight ok is more understandable than calling a anon function.
* if it has the handler, then call `get_handler_mod_and_props` which simply resolves the handler and returns `{HandlerMod, HandlerProps}` (basically change `do_with_handler` to return these) 
* now you have all you need, then directly call `update_metadata_with_handler/6`

Personally I read this code much easier, all function called are purposeful and i know what they are doing, and the main logic is is closer to its actions, I don't need to scroll back and forward to understand.

kazoo_documents/src/kzd_call_recordings.erl
```diff
@@ -35,6 +35,7 @@
 -export([start/1, start/2, set_start/2]).
 -export([to/1, to/2, set_to/2]).
 -export([url/1, url/2, set_url/2]).
+-export([key_user_metadata/0, user_metadata/1, user_metadata/2, set_user_metadata/2]).
```
icehess: How about `path_user_metadata`? This is more consistent with the rest of `kazoo_document` modules.

icehess (changes requested):

lazedo:
why the need for update_metadata and update_attachment_metadata in the first place ?
why not extend the put_object with optional metadata which gets translated on `s3` as `tags` and sent to erlcloud ?

kazoo_data/src/kzs_attachments.erl
danielfinke: Yeah, I was originally hoping to be a bit more ambitious with this and refactor out the common parts of `fetch_attachment`, `stream_attachment`, `delete_attachment`, and `update_metadata`. Simplified this for now as requested

danielfinke:
@lazedo `put_object` has also been extended to add the tags. Those tags are added when the object is first uploaded based on configured defaults.
`update_metadata` is used for retroactive changes to the metadata (see https://github.com/2600hz/kazoo-crossbar/pull/386) without requiring that the whole object is re-uploaded every time.

lazedo:
@danielfinke but why is this being added (retroactive changes) ? it seems we're trying to use the attachment metadata as data layer.

danielfinke:
@lazedo Verizon needs to be able to modify the importance of a recording after the fact (see https://2600hz-commercial.atlassian.net/browse/KZOO-273 and https://2600hz-commercial.atlassian.net/browse/KZOO-323). The `"important"` tag affects the retention policy for objects in the S3 call recordings bucket. I don't know what Verizon's business reasons are for marking a recording as important or not though.

lazedo:
thanks for the pointers @danielfinke , that explains it if they are using it that way (mark a recording not to be deleted in case there's a court order or something)

danielfinke:
@icehess or @lazedo: anything further needed here?

PR #541, opened 2022-11-28
KZOO-324 : Adding possibility to configure default custom application variables to appear on CDRs

sanjeevie6353:
KZOO-324 : Adding possibility to configure default custom application variables to appear on CDRs

Previously, the CDRs were printed with **custom application variables** only when the user has assigned them. 

This new feature will allow to define a set of  **custom application variables** in a system config. These defined sys.config values will always be available in the CDRs.


kazoo_call/src/kapps_call.erl
```diff
@@ -2045,3 +2046,47 @@ set_dynamic_cfwd(JObj, Call) ->
...
+    end.
+
+%%------------------------------------------------------------------------------
+%% @doc generates a key-value list to add as CAVs
+%% @end
+%%------------------------------------------------------------------------------
+build_default_cavs([], _UserDoc, CAVs) ->
+    CAVs;
+
+build_default_cavs([[ParentKey,ChildKey]| REST], UserDoc, CAVs) ->
```
jamesaimonetti: I'm not sure about this construction.
What are ParentKey and ChildKey supposed to look like?

jamesaimonetti: If you have a user doc with `{"foo":{ "bar": {"baz":123}}}` and you want "foo.bar.baz" set as a CAV, what would Parent/Child be here?

jamesaimonetti: My thinking is, just use a list of paths (a path = ["foo", "bar", "baz"] so [ ["foo", "bar", "baz"], ["a", "b"], ["enabled"] ] would be the list), and if `kz_json:get_value(Path, UserJObj)` is undefined, move on; if a a value is found, `kz_json:set_value(Path, Value, CAVs)` it.

kazoo_endpoint/src/kz_endpoints_v5.erl
```diff
@@ -14,10 +14,26 @@
...
         {ok, Endpoints} -> Endpoints;
         Error -> Error
     end.
+
+%%------------------------------------------------------------------------------
+%% @doc Will add default custom application variables on to the Call record if
+%% the keys defined in the sysconfig are available in the user doc
+%% @end
+%%------------------------------------------------------------------------------
+maybe_add_cavs_to_call(OwnerId, Data, Call) ->
```
jamesaimonetti: I think this work should be done in kazoo_directory (like kz_directory_user, kz_directory_device, etc).
These CAVs should be present in the directory XML sent to FS for the endpoint.

kazoo_call/src/kapps_call.erl
sanjeevie6353: This logic has been changed to implement `kz_json:get_value(Path, UserJObj)` and ` kz_json:set_value(Path, Value, CAVs)` logic

kazoo_endpoint/src/kz_endpoints_v5.erl
sanjeevie6353: The logic has been moved to `kz_directory_device`

kazoo_directory/src/kz_directory_device.erl
```diff
@@ -442,6 +443,46 @@ generate_routing_ccvs(Endpoint, _Options, Module) ->
...
+    CAVs;
+
+build_default_cavs([Path| REST], UserDoc, CAVs) ->
+    case kz_json:get_value(Path, UserDoc) of
+        'undefined' -> build_default_cavs(REST, UserDoc, CAVs);
+        CAVValue ->
+            CAVKey = lists:last(Path),
+            build_default_cavs(REST, UserDoc, kz_json:set_value(<<"cav_",CAVKey/binary>>, CAVValue, CAVs))
+    end.
+
```
lazedo, line 485: this shouldn't be done here
1.  `directory` should use `endpoint` data for device/user
2. these cavs should be set on kz_endpoint_v5 merge
3. default cavs should be configured for account/user/device
4. `Custom-Application-Vars` should be a `get_ne_json` on the endpoint (no defaults)

lazedo (changes requested):

kazoo_directory/src/kz_directory_device.erl
lazedo: also, when building the default cavs on kz_endpoint merge, don't use last part of path for the name of the variable, use an explicit mapping value {[<<"my-container">>, <<"my-variable-name">>], [<<"the">>, <<"path">>, <<"to">>, <<"value">>]}. build the values for account/user/device and merge them, set the value if the merged value is not empty

sanjeevie6353:
Closing the PR as the change on this PR is already merged onto master.

PR #547, opened 2022-11-30
spec `kz_json` `foldl/3`/`foldr/3` as polymorphic

danielfinke:
- also remove redundant unions in `kazoo_json.hrl`


kazoo_stdlib/include/kazoo_json.hrl
```diff
@@ -19,7 +19,7 @@
                               | json_array().
 
 -type non_null_json_term() :: non_object_json_term()
-                            | object() | ?EMPTY_JSON_OBJECT.
```
jamesaimonetti, line 22: I think we needed this because PropEr can't create a base-case for an object:

```
Error: The typeserver encountered an error: {no_base_case,{kz_json_generators,type,object,0}}.
```

jamesaimonetti:
```
      kz_json_tests: proper_test_ (prop_test_object_gen)...
Error: The typeserver encountered an error: {no_base_case,{kz_json_generators,type,object,0}}.
*failed*
in function kz_json_tests:'-proper_test_/0-fun-1-'/1 (test/kz_json_tests.erl, line 36)
in call from eunit_test:run_testfun/1 (eunit_test.erl, line 71)
in call from eunit_proc:run_test/1 (eunit_proc.erl, line 522)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 347)
in call from eunit_proc:handle_test/2 (eunit_proc.erl, line 505)
in call from eunit_proc:tests_inorder/3 (eunit_proc.erl, line 447)
in call from eunit_proc:with_timeout/3 (eunit_proc.erl, line 337)
in call from eunit_proc:run_group/2 (eunit_proc.erl, line 561)
**error:{assert,
    [{module,kz_json_tests},
     {line,36},
     {expression,
         "proper : quickcheck ( ? MODULE : F ( ) , [ { to_file , user } , { numtests , 1000 } ] )"},
     {expected,true},
     {not_boolean,
         {error,
             {typeserver,
                 {no_base_case,{kz_json_generators,type,object,0}}}}}]}
  output:<<"">>
```

danielfinke: Ah yeah, this is because of the `proper_transformer` parse transform: https://proper-testing.github.io/apidocs/proper_typeserver.html
> ?FORALLs may contain references to self-recursive or mutually recursive native types, so long as each type in the hierarchy has a clear base case. Currently, PropEr requires that the toplevel of any recursive type declaration is either a (maybe empty) list or a union containing at least one choice that doesn't reference the type directly (it may, however, reference any of the types that are mutually recursive with it).

```erlang
-ifdef(PROPER).
-type object() :: ?JSON_WRAPPER(json_ne_proplist()) |
                  %% Explicit PropEr base case
                  empty_object().
-else.
-type object() :: ?JSON_WRAPPER(json_proplist()) | map().
%% Denotes JSON object Erlang representation, {@link json_proplist} wrapped in `{}'.
-endif.
```
does the job

PR #549, opened 2022-12-06
KCRO-164: Add function to convert Crossbar objects into AMQP objects

harenson:
Dependency for https://github.com/2600hz/kazoo-crossbar/pull/399


kazoo_stdlib/src/kz_json.erl
```diff
@@ -2023,6 +2024,36 @@ search_replace_format({Old, New, Formatter}, JObj) when is_function(Formatter, 1
...
+%% each word (separated by -/_) with a single exception, "id" should be returned
+%% as "ID".
+%%
+%% Example:
+%%  input:  {[{<<"three_words_key">>, <<"value">>}]}
+%%  output: {[{<<"Three-Words-Key">>, <<"value">>}]}
+%% @end
+%%------------------------------------------------------------------------------
+%% {
+-spec normalize_amqp(kz_json:object()) -> kz_json:object().
```
jamesaimonetti: I thought we talked about putting this in kz_api? That module might be more appropriate since this is AMQP JSON specific

kazoo_stdlib/src/kz_json.erl
```diff
@@ -2023,6 +2024,36 @@ search_replace_format({Old, New, Formatter}, JObj) when is_function(Formatter, 1
...
+normalize_amqp(JObj) ->
+    foldl(fun fold_normalize_amqp/3, new(), JObj).
+
+-spec fold_normalize_amqp(key(), json_term(), kz_json:object()) -> kz_json:object().
+fold_normalize_amqp(Key, ?JSON_WRAPPER(_)=JObj, Acc) ->
+    %% Value is a nested object. Recursive normalization.
+    kz_json:set_value(normalize_amqp_key(Key), normalize_amqp(JObj), Acc);
```
jamesaimonetti: Just normalize the key, nested keys don't necessarily conform to Upper-Case-Var naming

kazoo_stdlib/src/kz_json.erl
harenson: Moved into _kz_api.erl_ module.

kazoo_stdlib/src/kz_json.erl
harenson: By default, it will only normalize root level keys.

lazedo:
what's the need for this pr @harenson ?

harenson:
@lazedo it is needed for completing KCRO-164 + KPUS-4 feature request. _users_ and _devices_ APIs are going to be able to receive an "arbitrary" payload, and this PR's changes are needed to normalize that payload before sending it to _pusher_ application via _kapi_pusher.erl_ module.

lazedo:
@harenson that doesn't make any sense. `arbitrary` means `any` payload `not` `any payload that is normalized in a very opiniated form`

lazedo:
i still don't see a reason for this

harenson:
@lazedo That is why I said "normalized" (notice the quotes). Not all the content within the payload is arbitrary.
 
 This normalization process was requested verbally by @k-anderson for this new feature.

harenson:
@lazedo @jamesaimonetti What change do you suggest to get this PR ready for review/merge or close?

lazedo:
@harenson i still don't see a reason for this, but not blocking anything here

kazoo_amqp/test/kz_api_tests.erl
```diff
@@ -47,6 +48,142 @@ get_event_type_test_() ->
     ,?_assertEqual({<<"call">>,<<"CHANNEL_CONNECTED">>}, kz_api:event_type(kz_json:from_list([EventCategory,EventName])))
     ].
 
+normalize_amqp_test_() ->
+    Test = fun kz_api:normalize_amqp/1,
+    Test2 = fun kz_api:normalize_amqp/2,
+    ToJSON = fun(Prop) -> kz_json:from_list_recursive(Prop) end,
```
jamesaimonetti: `ToJSON = fun kz_json:from_list_recursive/1` is sufficient, no need to create a fun for this

lazedo:
@harenson can you explain why is this needed ? it seems like you're trying to `reverse engineer` something, when the problem may be in the design of the payloads where this is supposed to be used

harenson:
@lazedo This new function is used to convert Crossbar payloads (lowercase words separated by an underscore character) into AMQP payloads (First letter of each work in upper case, with words separated by a hyphen character).

lazedo:
@harenson i didn't asked what it does, i asked if you could explain `why` is this needed.it seems like you're trying to reverse engineer something, when the problem may be in the design of the payloads where this is supposed to be used. where is this going to be used and `why` ?

harenson:
@lazedo `/v2/{ACCOUNT_ID}/{ENDPOINT_TYPE}/{ENDPOINT_ID}/pushes` Crossbar [API schema](https://github.com/2600hz/kazoo-crossbar/blob/6ffef3a8603d4d55253aa0950f0ae6221702e3f0/priv/couchdb/schemas/endpoint.push.json) and [_kapi.pusher.endpoint_push_req_ schema](https://github.com/2600hz/kazoo-core/blob/master/kazoo_amqp/priv/couchdb/schemas/kapi.pusher.endpoint_push_req.json) share almost the same schema, they differ in the words separator, capitalization, and also _id_ in AMQP payloads is expected to be converted into _ID_.

For example, if `/pushes` Crossbar API receives a payload like:
```
{
  "data": {
    "apns": {
      "alert": {
        "subtitle_key": "some subtitle",
        "subtitle_params": [
          "params",
          "here"
        ]
      },
      "push_type": "alert",
      "thread_id": "some thread ID",
      "topic": "some topic"
    }
  }
}
```

`/pushes` Crossbar API needs to convert the payload into:
```
{
  "APNs": {
    "Alert": {
      "Subtitle-Key": "some subtitle",
      "Subtitle-Params": [
        "params",
        "here"
      ]
    },
    "Push-Type": "alert",
    "Thread-ID": "some thread ID",
    "Topic": "some topic"
  }
}
```

before sending it to Pusher via AMQP, because that is what the Pusher's schema expect.

lazedo:
@harenson this doesn't make any sense to me at so many levels that i find it very hard to describe it. Crossbar [API schema](https://github.com/2600hz/kazoo-crossbar/blob/6ffef3a8603d4d55253aa0950f0ae6221702e3f0/priv/couchdb/schemas/endpoint.push.json) and [kapi.pusher.endpoint_push_req schema](https://github.com/2600hz/kazoo-core/blob/master/kazoo_amqp/priv/couchdb/schemas/kapi.pusher.endpoint_push_req.json) are `imho` wrong from the start.
anyway, i do not agree with this and unfortunately the changes to pusher were accepted. i think we should go back to those pusher changes and do it properly since its still only on master.

another note, `pushes` seems very odd, `push` (action) seems more appropriate like `/.../devices/id/notify` and not `/.../devices/id/notifications`

as for the `review`, if we want to proceed with this approach, this code should be put where its used and not go into core.

SUB-Title-Text -> normalize -> sub_title_text -> normalize_amqp -> Sub-Title-Text

jamesaimonetti:
@lazedo naming is hard! we settled on "pushes" as it is a noun too "the push to get this PR merged" and putting actions in the URI isn't typically what we want.

harenson:
Functionality moved into Crossbar application: https://github.com/2600hz/kazoo-crossbar/pull/399 PR.

PR #557, opened 2022-12-14
KZOO-336 add channel count and endpoint merge limits

lazedo:


lazedo:
spellcheck was trying to check a deleted file

PR #558, opened 2022-12-19
KCRO-149: Implementation of Websites API

Navoda626:
KCRO-149: Implementation of Websites API

Implementation of websites API to facilitate adding websites using the comm.land. So an admin and users would be able to add custom websites with a logo.

A new document with the type 'website' is created in the database for every website added. 'name' and the 'web_url' are mandatory parameters for a website document and there is 'users' parameter which holds the list of users whom the website is assigned. And there is a 'account_wide' parameter works as a flag to assign the website to every user in that account.

Create, update, delete and read APIs implemented for the websites.

Additionally a binary logo can be updated to a website document and fetched from a website document.


kazoo_stdlib/include/kz_databases.hrl
```diff
@@ -51,6 +52,7 @@
                        ,?KZ_ACDC_DB
                        ,?KZ_OAUTH_DB
                        ,?KZ_FUNCTIONS_DB
+                       ,?KZ_WEBSITES_DB
```
jamesaimonetti: Why this change? I don't see the need for a `websites` db nor do I see it used in the Crossbar PR?

kazoo_data/src/kzs_util.erl
```diff
@@ -92,6 +92,7 @@ db_classification(?KZ_FAXES_DB) -> 'aggregate';
 db_classification(?KZ_ACDC_DB) -> 'aggregate';
 db_classification(?KZ_SERVICES_DB) -> 'aggregate';
 db_classification(?KZ_PORT_REQUESTS_DB) -> 'aggregate';
+db_classification(?KZ_WEBSITES_DB) -> 'aggregate';
```
jamesaimonetti: Seems unnecessary

jamesaimonetti (changes requested):

kazoo_stdlib/include/kz_databases.hrl
Navoda626: Yes this was initially added. This is not necessary. Will resolve this

kazoo_data/src/kzs_util.erl
Navoda626: Yes this is not necessary. Will resolve this

PR #559, opened 2022-12-21
[5.1] PS-111: add support for cleanup of extended MODB databases (#554)

nkarunarathne:
Previously when an account was deleted extended MODB databases were not deleted.
To solve that now deletes extended MODB databases with other modbs when deleting an account.

- new database classification added to db_classification called 'type_modb' (matches something like <<"account/e1/82/5a77626f4..........-202210-whatever">>)
- create a new function to get modbs and other types of modbs of a given account
- added functionality to format extended MODB databases names


sanjeevie6353:
Checked the backport PR and this has the same changes that pushed onto the master... 
https://github.com/2600hz/kazoo-core/pull/554

PR #579, opened 2023-02-09
[5.1] KCRO-149: Websites API - Website view change to fix duplicating website documents

Navoda626:
Earlier there was a bug where duplicating website documents in the REST response when account_wide=true (meaning allowed for all users in the account) and users=[user_id] (meaning allowed for specified user ids in the list). This scenario is not likely since if account_wide=true then no need of mentioning user ids in the users array.

With this change in couchdb view, only emit to the account_wide key once when account_wide is true; if not, enumerate the users and emit the user as key.

(cherry picked from commit 1be111815c80677abc4a9e3109648bac4f356b0a)


nkarunarathne:
This looks good

PR #585, opened 2023-02-16
KCRO-170: Add feature flag 'getstream' if the user doc.chat.getstream.enabled = true

Navoda626:
KCRO-170: Add feature flag 'getstream' if the user doc.chat.getstream.enabled=true

Previously there was no feature for user document to enable/disable getstream chat workspace.

Now, getstream chat workspace will be enabled/disabled userwise by an admin using the mosterUI. With the change If a specific user is been enabled by a admin then 'getstream' feature flag is enabled for that specific user and it will show on the user fetch response


nkarunarathne:
Changes look good

PR #586, opened 2023-02-16
KCRO-168: Validating the getstream feature enabled in user document

Navoda626:
KCRO-168: Validating the getstream feature enabled in user document

Earlier when the getstream API PUT /accounts/{ACCOUNT_ID}/users/{USER_ID}/getstream/connect is called user is connected to getstream API without validating whether the user has getstream feature (chat.getstream.emabled=true).

Now, added chat.getstream.emabled parameter is the user document schema in crossbar and in core added kzd functions related to chat parameters


nkarunarathne:
Changes look good

lazedo (changes requested):
why are these being added here ?
kzd_users should not have app specifics
maybe add a kzd_chat and kzd_getstream each dealing with properties of their domain

Navoda626:
> why are these being added here ? kzd_users should not have app specifics maybe add a kzd_chat and kzd_getstream each dealing with properties of their domain

@lazedo This chat.getstream.enabled parameter is added to the user document schema as a feature that a user has. Is it still  need to be moved to other module?

Navoda626:
Accessor functions are moved to a kzd_chat.erl module since they are application specific parameters

Navoda626:
Moved the required accessor functions to cb_getstream.erl module. So closing this PR

PR #587, opened 2023-02-16
KZOO-346: update the kazoo-core services quantity handlers to emit quantities for chat.getstream.enabled

Navoda626:
KZOO-346: update the kazoo-core services quantity handlers to emit quantities for chat.getstream.enabled

Charge $1 per user when the getstream chat workspace is enabled (user document will have chat.getstream.enable=true) for a user by an Admin


nkarunarathne:
Changes look good

kazoo_services/src/kz_services_quantities.erl
```diff
@@ -650,6 +651,13 @@ is_first_transition_to_state(JObj) ->
                                         ),
     length(StateTransitions) =:= 1.
 
+-spec calculate_chat_updates(kz_json:object(), kz_term:proplist()) -> kz_term:proplist().
+calculate_chat_updates(JObj, Updates) ->
+    case kz_json:is_true([<<"chat">>, <<"getstream">>, <<"enabled">>], JObj, 'false') of
+        'true' -> [{[<<"chat">>, <<"getstream">>], 1} | Updates];
```
jamesaimonetti, line 661: Is the `1` for one dollar? or for `1` quantity?

Navoda626: As I understand it is 1 quantity

Navoda626: But in the story of the ticket it said $1 charge per user. https://2600hz-commercial.atlassian.net/browse/MUDA-13

So I'm not sure how this quantity convert into $. I will check on that

Navoda626: This 1 is for 1 quantity and conversion to currency is done in kz_currency module

jamesaimonetti: update here, right? should be pvt_chat now, yes?

Navoda626: Yes. The user document field is updated

PR #589, opened 2023-02-16
[5.1] KCRO-168: Validating the getstream feature enabled in user document

Navoda626:
[5.1] KCRO-168: Validating the getstream feature enabled in user document

Earlier when the getstream API PUT /accounts/{ACCOUNT_ID}/users/{USER_ID}/getstream/connect is called user is connected to getstream API without validating whether the user has getstream feature (chat.getstream.emabled=true).

Now, added chat.getstream.emabled parameter is the user document schema in crossbar and in core added kzd functions related to chat parameters

(cherry picked from commit 82b93f4e74133fbbac77f6fec57cafdd30a792ff)


Navoda626:

PR #590, opened 2023-02-16
[5.1] KCRO-170: Add feature flag 'getstream' if the user doc.chat.getstream.enabled = true

Navoda626:
[5.1] KCRO-170: Add feature flag 'getstream' if the user doc.chat.getstream.enabled=true

Previously there was no feature for user document to enable/disable getstream chat workspace.

Now, getstream chat workspace will be enabled/disabled userwise by an admin using the mosterUI. With the change If a specific user is been enabled by a admin then 'getstream' feature flag is enabled for that specific user and it will show on the user fetch response

(cherry picked from commit 117594293a434d894d57d04b2a5baa52117c290b)


Navoda626:

PR #591, opened 2023-02-16
[5.1] KZOO-346: update the kazoo-core services quantity handlers to emit quantities for chat.getstream.enabled

Navoda626:
[5.1] KZOO-346: update the kazoo-core services quantity handlers to emit quantities for chat.getstream.enabled

Charge $1 per user when the getstream chat workspace is enabled (user document will have chat.getstream.enable=true) for a user by an Admin

(cherry picked from commit 97416362e01be9c711144ce6685de8a6035471db)


Navoda626:

PR #596, opened 2023-02-23
KCRO-173: Add an optional field open_in_browser to the website document

Navoda626:
KCRO-173: Add an optional field open_in_browser to the website document

Adding a boolean optional parameter 'open_in_browser' default set as 'false' for the website document so that the admin or the user adding the website can specify that url to be opened specifically in browser other than the commland application (Since some websites doesn't open as expected in commland application, so those need to be opened in browser).

Website views are modified to fetch the new field 'open_in_browser' and added kzd_websites.erl new functions related to the new fields


sanjeevie6353:
The PR seems to ok to me. The CircleCI fails are only due to dependency with https://github.com/2600hz/kazoo-crossbar/pull/429 PR.

PR #598, opened 2023-02-23
[5.1] FREE-56: add the number that is fowarding the call to ccvs

icehess:
Adding the the caller id number that is forwarding the call so we can process it later on the call.
For example we use this value to set `diversion` header in `cf_resource`.


kazoo_endpoint/src/kz_endpoint_v5.erl
```diff
@@ -1473,6 +1473,7 @@ maybe_set_call_forward({Endpoint, Call, CallFwd, CCVs}, Number) ->
                         ,{<<"Authorizing-Type">>, <<"device">>}
                         ,{<<"Authorizing-ID">>, kz_doc:id(Endpoint)}
                         ,{<<"Call-Forward-From">>, kapps_call:inception_type(Call)}
+                        ,{<<"Call-Forward-Number">>, kapps_call:caller_id_number(Call)}
```
lazedo: why changing here ? if really needed, the value is wrong, it should be the called number.
i would drop this change

icehess:
Backport PRs:

- master: https://github.com/2600hz/kazoo-core/pull/600
- 5.0: https://github.com/2600hz/kazoo-core/pull/599

PR #601, opened 2023-02-24
[5.1] FREE-56: add SIP-Diversions to offnet_req

icehess:


lazedo:
closing with f392ef81f0f577c2ad892ee04a615bb972339e16

PR #602, opened 2023-02-24
[5.0] FREE-56: add SIP-Diversions to offnet_req

icehess:


lazedo:
closing with b84872d38ffcadfddda866b5cf35f7076f2e4442

PR #603, opened 2023-02-24
[master] FREE-56: add SIP-Diversions to offnet_req

icehess:


icehess:
Backport PRs:
- 5.1: https://github.com/2600hz/kazoo-core/pull/601
- 5.0: https://github.com/2600hz/kazoo-core/pull/602

PR #608, opened 2023-03-01
Defer parsing URL to OTP uri_string

jamesaimonetti:
Previously, there was a bug in the implementation where the username/password was separated by a ':' and the host came after the '@' which violates the expectation of {user}@{pass}:{host}.

In OTP 21, a module uri_string is introduced with parsing abilities. This PR updates the tests to properly test the structure of a URI with basic auth and use uri_string for parsing the binary URI.


jamesaimonetti:
Test updates: https://github.com/2600hz/kazoo-properly/pull/130

PR #609, opened 2023-03-03
KZOO-303: add password requirements enforcer

icehess:
This will add an configurable password requirements enforcer to `kz_users` validation.

The default is settings is enabling the the enforcers. The enforcer is a list of regular experssion to check the requirements of the password.

The settings are configurable per account, reseller and system_config.


icehess:
Required PR:
- https://github.com/2600hz/kazoo-core/pull/609
- https://github.com/2600hz/kazoo-crossbar/pull/434

Properly test:
- https://github.com/2600hz/kazoo-properly/pull/143

icehess:
CI is failing because doc steps required this circular deps PR: https://github.com/2600hz/kazoo-crossbar/pull/434

kazoo_auth/src/kz_auth_password.erl
```diff
@@ -0,0 +1,171 @@
...
+         )
+       ).
+
+-define(SHOULD_PREVENT_REUSE(AccountId)
+       ,kz_term:is_true(
+          kapps_account_config:get_with_strategy(<<"global_merge">>, AccountId, ?MOD_CAT, <<"should_prevent_reuse">>, 'false')
+         )
+       ).
+
+-define(STRENGTH_REGESEX(AccountId)
```
jamesaimonetti: what is rege-sex? :p

icehess: because i can't spell :))

kazoo_auth/src/kz_auth_password.erl
```diff
@@ -0,0 +1,171 @@
...
+-spec maybe_enforce(kz_term:api_ne_binary(), kz_term:ne_binary(), boolean()) -> enforce_resp().
+maybe_enforce(_AccountId, Password, 'false') ->
+    {'valid', Password};
+maybe_enforce(AccountId, Password, 'true') ->
+    validate_password(AccountId, Password).
+
+-type valid_acc() :: {kz_term:ne_binary(), kz_term:ne_binaries()}.
+-spec do_validate_password(kz_term:ne_binary(), kz_json:object(), valid_acc()) -> valid_acc().
+do_validate_password(Name, StrengthRegex, {Password, Errors}) ->
+    Regex = kz_json:get_ne_binary_value(<<"regex">>, StrengthRegex),
```
jamesaimonetti, line 155: Can regex ever be `'undefined'`? Just curious why `re:run` below needs the is_ne_binary before it?

Second question is why the `try` below for the regex?

kazoo_auth/src/kz_auth_password.erl
```diff
@@ -0,0 +1,171 @@
...
+            lager:error("empty password validation regex, regex: ~s", [Name]),
+            {Password, Errors};
+        {match, _} ->
+            {Password, Errors};
+        'nomatch' ->
+            lager:debug("password validation failed: ~s", [Message]),
+            {Password, [Message | Errors]}
+    catch _E:_T:_ST ->
+            lager:error("bad password validation regex, regex(~s): ~p", [Name, Regex]),
+            {Password, Errors}
```
jamesaimonetti: Could a "well crafted" password cause a regex match to crash? Should a bad regex rule, causing a crash, allow passwords to continue on their journey?

kazoo_auth/src/kz_auth_password.erl
icehess: It is not protecting the sane default config, it is protecting from people configuring things badly. Same for try, if people use a regex which is not correct and cause re:run to crash.

kazoo_auth/src/kz_auth_password.erl
icehess: philosophically, maybe `¯\_(ツ)_/¯`.

You can claim the same if you "well craft" an invalid password but still passes the regex? or craft a password that utilized a a bug in re:run to cause it to crash/pass the regex?

We can fail and return error, i don't have any preference.

icehess: If we hard coded the regex, i wouldn't use try/catch, but since this is configurable per account, it is better to anticipate a bad regex.

jamesaimonetti: I think, if re:run crashes, the password should not be accepted, we should log loudly to double-check the regex, and make life hell for the sysadmin who messed it up!

As-is, if I mess up my regex but assume it is working as expected, then find out later its not cause my customers got hacked because of weak passwords... i'm a gonna be in a pickle

PR #610, opened 2023-03-03
KZOO-304: Implementation sanitizer for HTML input

dhuynh-Kazoo:


kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,383 @@
...
+  re:replace(HTML, "\\<\\!\\[CDATA\\[", "", [global, {return, binary}]).
+
+-spec parse(binary() | string()) -> html_node().
+parse(HTML) when is_list(HTML)->
+  parse(list_to_binary(HTML));
+parse(HTML) ->
+  HTML1 = case is_binary(HTML) of
+              true -> HTML;
+              false -> list_to_binary(HTML)
+          end,
```
icehess: I think this extra. you can omit it, you already convert list to binary in previous func clause.

But alos to confirm you are operating on binary, then pattern match in the header, and let code crash as soon as possible.

```suggestion
parse(<<HTML/binary>>) ->
```

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,383 @@
...
+  ScrubAttributes = lists:map(fun(Attr) ->
+                                ScrubAttr(Attr)
+                              end, Attributes),
+  {<<"style">>, ScrubAttributes, [scrub_css(Text)]};
+
+scrub({Tag, Attrs, Data}) ->
+  case lists:member(Tag, proplists:get_keys(?ALLOW_TAG_WITH_NORMAL_ATTRIBUTES))
+        orelse
+       lists:member(Tag, proplists:get_keys(?ALLOW_TAG_WITH_URI_ATTRIBUTES))
+  of
```
icehess: We usually drop orelse/andalso/and/or to new line like this:

```suggestion
    case lists:member(Tag, proplists:get_keys(?ALLOW_TAG_WITH_NORMAL_ATTRIBUTES))
        orelse lists:member(Tag, proplists:get_keys(?ALLOW_TAG_WITH_URI_ATTRIBUTES))
    of
```

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,383 @@
...
+-define(HTTP_LIKE_SCHEME, lists:concat(["(?<scheme>.+?)", "(", ?PROTOCOL_SEPARATOR, ")//"])).
+-define(OTHER_SCHEMES, lists:concat(["(?<other_schemes>mailto)","(", ?PROTOCOL_SEPARATOR,")"])).
+-define(SCHEME_CAPTURE, lists:concat(["(",?HTTP_LIKE_SCHEME , ")", "|", "(", ?OTHER_SCHEMES, ")"])).
+
+%% API
+-export([apply/1,
+         parse/1,
+         traverse/1,
+         scrub_css/1,
+         to_html/1]).
```
icehess: We always format list as comma-first:
```suggestion
-export([apply/1
        ,parse/1
        ,traverse/1
        ,scrub_css/1
        ,to_html/1
        ]
       ).
```

icehess (changes requested):
In first pass this looks good to me.
As you notice we are a bit strict about code formatting.
There are some handy way to run the formatter. The simplest way is to make your change, commit (not push yet) and run:

```
make fmt
```

This will format your codes, you need to commit those changes again.

We use 4 space (no tab at all)

Also we use comma-first list, tuple at etc... like this:

```
-export([apply/1
        ,parse/1
        ,traverse/1
        ,scrub_css/1
        ,to_html/1
        ]
       ).
```

icehess (changes requested):
@dhuynh-Kazoo Also please always make sure you have the ticket number in the PR title;
`KZOO-304: Implementation sanitizer for HTML input`

And also please always add some description to your first commit and also add description to PR.

kazoo_web/src/kz_html_sanitizer.erl
jamesaimonetti: Also use the `props` module in kazoo_stdlib instead of `proplists` when possible

kazoo_web/src/kz_html_sanitizer.erl
dhuynh-Kazoo: Formatted

kazoo_web/src/kz_html_sanitizer.erl
dhuynh-Kazoo: Replaced with props

dhuynh-Kazoo: formatted case...of

kazoo_web/src/kz_html_sanitizer.erl
dhuynh-Kazoo: Ongoing, I am checking the fields in the account then I'll standardize the input type

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,344 @@
...
+    re:replace(HTML2, "(\&gt\;|>)(\ +)(" ++ ?REPLACEMENT_TAB ++ ")(\t+)(\&lt\;|<)",
+               "\\1\\4\\5", [{return, binary}, global]).
+
+-spec traverse(list() | html_node()) -> html_node() | html_attr().
+traverse([]) -> [];
+traverse([Head|Tail]) ->
+    Head_1 = traverse(Head),
+    Head_2 = collapse_list(Head_1),
+    Tail_1 = traverse(Tail),
+    lists:flatten([Head_2] ++ Tail_1);
```
jamesaimonetti: `[Head_2 | Tail_1]` should be equivalent and not rely on the append operator

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,344 @@
...
+traverse({Token, Children}) ->
+    Children_1 = traverse(Children),
+    Children_2 = collapse_list(Children_1),
+    scrub({Token, Children_2});
+traverse(Text) when is_binary(Text) ->
+    scrub(Text);
+traverse(Anything) ->
+    Anything.
+
+scrub({<<"script">>, _, _}) -> [];
```
jamesaimonetti: Empty list seems like an odd choice. Why not use `undefined` and filter it out in traverse/1?

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,344 @@
...
+        {match, _} -> Src;
+        nomatch -> []
+    end;
+scrub_reference_src(Src) ->
+    case capture_scheme(Src) of
+        true -> Src;
+        false -> []
+    end.
+
+collapse_list(List) ->
```
jamesaimonetti: In general, use pattern-matching in the function clause when possible, then in case statements.

```erlang
collapse_list([X]) -> X;
collapse_list(L) -> L.
```

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,344 @@
...
+after_to_html(HTML) ->
+    HTML1 = re:replace(HTML, "(\ ?" ++ ?REPLACEMENT_LINEBREAK ++ " )(\r?\n)",
+                       "\\2", [{return, binary}, global]),
+    HTML2 = re:replace(HTML1, "(\&gt\;|>)(\ +)(" ++ ?REPLACEMENT_SPACE ++ ")(\ +)(\&lt\;|<)",
+                       "\\1\\4\\5", [{return, binary}, global]),
+    re:replace(HTML2, "(\&gt\;|>)(\ +)(" ++ ?REPLACEMENT_TAB ++ ")(\t+)(\&lt\;|<)",
+               "\\1\\4\\5", [{return, binary}, global]).
+
+-spec traverse(list() | html_node()) -> html_node() | html_attr().
+traverse([]) -> [];
```
jamesaimonetti: Seems like a list comprehension to map a function of the HTML node would be easier to read?

```erlang
traverse(HTMLNodes) ->
    [Scrubbed || HTMLNode <- HTMLNodes, Scrubbed <- [handle_node(HTMLNode)], Scrubbed =/= 'undefined'].
```

Something along these lines, then `handle_node/1`  becomes the traverse clauses below for handling the tuples. And if scrub returns 'undefined', you can filter scrubbed results.

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,344 @@
...
+    end.
+
+allowed_keyword(Val) when is_binary(Val) ->
+    lists:member(string:lowercase(Val), ?CSS_ALLOW_KEYWORDS);
+allowed_keyword(Val) -> allowed_keyword(list_to_binary(Val)).
+
+measured_unit(Val) ->
+    Res = re:run(Val,
+                 "\\A(#[0-9a-f]+|rgb\\(\\d+%?,\\d*%?,?\\d*%?\\)?|-?\\d{0,2}\\.?\\d{0,2}(cm|em|ex|in|mm|pc|pt|px|%|,|\\))?)\\z"),
+    case Res of
```
jamesaimonetti: `Res =/= 'nomatch'.` unless you plan to use the matched groups

dhuynh-Kazoo:
Good morning @jamesaimonetti and @icehess , could you please review the code again? I changed the implementation to check if the HTML contains XSS or not. The sanitizer function was removed.

kazoo_schemas/src/kz_json_schema_extensions.erl
```diff
@@ -157,6 +160,17 @@ extra_validation(_Key, _Value, State) ->
     lager:info("extra validation of ~s not handled for value ~p", [_Key, _Value]),
     State.
 
+announcement_validation(Value, State) ->
+    case kz_html_sanitizer:check_xss_items(Value) of
+        <<>> -> State;
+        Items ->
+            ErrorMsg = <<"Announcement contains cross-site scripting: ", Items/binary>>,
+            lager:debug("Announcement contains cross-site scripting ~s", [ErrorMsg]),
```
jamesaimonetti: log lines start with lowercase characters, as part of the full log line in syslog

kazoo_schemas/src/kz_json_schema_extensions.erl
```diff
@@ -157,6 +160,17 @@ extra_validation(_Key, _Value, State) ->
...
+            jesse_error:handle_data_invalid({'external_error',ErrorMsg}
+                                            ,Value
+                                            ,State)
```
jamesaimonetti: Drop the paren to the newline

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,169 @@
+%%%-------------------------------------------------------------------
+%%% @author dhuynh
+%%% @copyright (C) 2023, 2600hz
+%%% @doc
+%%% @end
+%%% Created : 03. Mar 2023 2:14 AM
+%%%-------------------------------------------------------------------
+-module(kz_html_sanitizer).
+
+-define(VALID_SCHEMES, ["http", "https", "mailto"]).
```
jamesaimonetti: Unless required to, use binaries instead of strings

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,169 @@
...
+-define(PROTOCOL_SEPARATOR, ":|(&#0*58)|(&#x70)|(&#x0*3a)|(%|&#37;)3A").
+-define(HTTP_LIKE_SCHEME, lists:concat(["(?<scheme>.+?)", "(", ?PROTOCOL_SEPARATOR, ")//"])).
```
jamesaimonetti: The compiler will flatten strings, no need to run lists:concat I think.

`-define(M, "foo" "bar" "baz").` will render as `-define(M, "foobarbaz").`

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,169 @@
...
+        false -> false;
+        true -> {true, {Attr, Src}}
+    end.
+
+find_xss_in_css(Text) ->
+    Text1 = re:replace(Text, "(\\/\\*|\\*\\/|<!--|-->)", " ", [{return, list}, global]),
+    Res =
+        case re:run(Text1,"([-\\w]+)\\s*:\\s*([^:;]*)",[global]) of
+            {match, List} ->
+                lists:filtermap(fun([{Pos1, Len1}, {Pos2, Len2},{Pos3, Len3}]) ->
```
jamesaimonetti: typically like to put this into a called function, to avoid large horizontal nesting.

dhuynh-Kazoo: Done

kazoo_web/src/kz_html_sanitizer.erl
dhuynh-Kazoo: Done

kazoo_web/src/kz_html_sanitizer.erl
dhuynh-Kazoo: Done

kazoo_schemas/src/kz_json_schema_extensions.erl
dhuynh-Kazoo: Done

kazoo_schemas/src/kz_json_schema_extensions.erl
dhuynh-Kazoo: Done

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+    case xss_items_handler(Node) of
+        true -> true;
+        false -> check_xss_items_in_nodes(Rest)
+    end.
+
+xss_items_handler({comment, _}) -> false;
+xss_items_handler({_, SubNode}) -> check_xss_items_in_nodes(SubNode);
+xss_items_handler({<<"script">>, _Attrs, _SubNode}) -> true;
+xss_items_handler({<<"style">>, _Attrs, Text}) ->
+    Res = lists:filtermap(fun(Elem) ->
```
jamesaimonetti: This could just be `Res = lists:filtermap(fun find_xx_in_css/1, Test),` no need for the anonymous function

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+
+%% @doc
+%% Check if the HTML input can contain Cross-Site Scripting items
+%% If it detects any XSS item, returns `true`. Otherwise, it continue traverse the next node
+%% NOTES: has_css_items/1 and get_css_items/1 do not use the same handler function
+%% Because has_css_items/1 stops and returns `true` when it meets the first XSS item
+%% get_css_items/1 has to traverse all HTML nodes to collect all XSS items
+%% @end
+-spec has_xss_items(binary()) -> boolean().
+has_xss_items(<<>>) -> false;
```
jamesaimonetti: Ticky atoms

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+get_xss_items([Node|Rest], Acc) ->
+    XssNode = get_xss_items_handler(Node),
+    get_xss_items(Rest, [XssNode|Acc]).
+
+get_xss_items_handler({comment, _}) -> [];
+get_xss_items_handler({_, SubNode}) -> get_xss_items(SubNode, []);
+get_xss_items_handler({<<"script">>, _Attrs, _SubNode} = Node) ->
+    {xss_script, kz_html_sanitizer:to_html(Node)};
+get_xss_items_handler({<<"style">>, _Attrs, Text} = Node) ->
+    Res = lists:filtermap(fun(Elem) ->
```
jamesaimonetti: same here, just pass `fun find_xss_in_css/1` instead

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+        case find_xss_in_attributes(Attrs) of
+            [] -> [];
+            XssAttr ->
+                {xss_attr, kz_html_sanitizer:to_html([{Tag, XssAttr, []}])}
+        end,
+    [X || X <- [XssAttrs, XssSubNode], X =/= []];
+get_xss_items_handler(<<_Node/binary>>) -> [].
+
+find_xss_in_attributes(Attrs) ->
+    lists:filtermap(fun(Attr) ->
```
jamesaimonetti: here too

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+get_xss_items(<<HTML/binary>>) ->
+    ParsedHTML = kz_html_sanitizer:simple_parse(HTML),
+    get_xss_items(ParsedHTML, []).
+
+get_xss_items([], Acc) -> lists:flatten(Acc);
+get_xss_items([Node|Rest], Acc) ->
+    XssNode = get_xss_items_handler(Node),
+    get_xss_items(Rest, [XssNode|Acc]).
+
+get_xss_items_handler({comment, _}) -> [];
```
jamesaimonetti: Can you explain why you chose `[]` as the return value here and in other clauses below?

You could try returning `'undefined'`; then instead of lists:flatten above you could `props:filter_undefined(Acc)`. Might be worth comparing run times of the two approaches.

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+        false -> false
+    end;
+find_xss_in_val(Attr, Src) ->
+    case not kz_html_sanitizer:validate_scheme(Src) of
+        false -> false;
+        true -> {true, {Attr, Src}}
+    end.
+
+find_xss_in_css(Text) ->
+    Text1 = re:replace(Text,
```
jamesaimonetti: Commas drop to newline:

```erlang
re:replace(Text
           ,"..."
```

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+                       "(\\/\\*|\\*\\/|<!--|-->)",
+                       " ",
+                       [{return, list}, global]),
```
jamesaimonetti: Matching parens should be vertically aligned, drop `),` down to align with `re:replace(` above

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+    Fun = fun([{Pos1, Len1}, {Pos2, Len2},{Pos3, Len3}]) ->
+                  SubMatched = string:substr(Text1, Pos1+1, Len1),
+                  _Property = string:substr(Text1, Pos2+1, Len2),
```
jamesaimonetti: seems unnecessary

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+                  Value = string:substr(Text1, Pos3+1, Len3),
+                  case find_xss_property_val(Value) of
+                      [] -> false;
+                      _ -> {true, SubMatched}
+                  end
+          end,
+    Res =
+        case re:run(Text1,"([-\\w]+)\\s*:\\s*([^:;]*)",[global]) of
```
jamesaimonetti: why not capture the values of the groups instead of the positions? Then you wouldn't need to do the string:substr above right?

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,207 @@
+%%%-------------------------------------------------------------------
+%%% @author dhuynh
+%%% @copyright (C) 2023, 2600hz
+%%% @doc
+%%% Check the HTML string contains the Cross Site Scripting (XSS)
+%%% 3 important elements will be checked
+%%* * some attributes and their values like href, src,...
+%%% * CSS properties and their values, formatted likes <word-word>: <word-word>
```
jamesaimonetti: html escape the `<` and `>` for edoc to pass

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
+%%%-------------------------------------------------------------------
+%%% @author dhuynh
+%%% @copyright (C) 2023, 2600hz
+%%% @doc
+%%% Check the HTML string contains the Cross Site Scripting (XSS)
+%%% 3 important elements will be checked
+%%* * some attributes and their values like href, src,...
+%%% * CSS properties and their values, formatted likes <word-word>: <word-word>
```
jamesaimonetti: escape here too

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+-include("kazoo_html_sanitizer_whitelist.hrl").
+
+%% API
+-export([has_xss_items/1
+        ,get_xss_items/1
+        ]).
+
+%% @doc
+%% Check if the HTML input can contain Cross-Site Scripting items
+%% If it detects any XSS item, returns `true`. Otherwise, it continue traverse the next node
```
jamesaimonetti: edoc doesn't seem to like the backticks here, maybe just remove

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,167 @@
...
+%%% Check the HTML string contains the Cross Site Scripting (XSS)
+%%% 3 important elements will be checked
+%%* * some attributes and their values like href, src,...
+%%% * CSS properties and their values, formatted likes <word-word>: <word-word>
+%%% * script tag
+%%% @end
+%%% Created : 03. Mar 2023 2:14 AM
+%%%-------------------------------------------------------------------
+-module(kz_html_xss).
+-include("kazoo_html_sanitizer_whitelist.hrl").
```
jamesaimonetti: include goes after export

kazoo_web/src/kz_html_xss.erl
icehess: Edoc treat some characters, to avoid that use Edoc code format.
It starts with a back tick ` and ENDS with a single quote

https://www.erlang.org/doc/apps/edoc/chapter.html#verbatim-quoting

kazoo_web/src/kz_html_xss.erl
icehess: Edoc treat some characters, to avoid that use Edoc code format.
It starts with a back tick ` and ENDS with a single quote

https://www.erlang.org/doc/apps/edoc/chapter.html#verbatim-quoting

kazoo_web/src/kz_html_xss.erl
dhuynh-Kazoo: Hi @jamesaimonetti , my idea is that the function may return a prop list like `[{xss_script, Items}, {xss_attr, Items}, {xss_script, Items}]` for the type of XSS that detected in HTML. If we do not find anything, return the `[]`.

kazoo_web/src/kz_html_xss.erl
dhuynh-Kazoo: Hi @jamesaimonetti, I deleted this case because the below one already handles it. But what do you mean by `ticky atoms`?

kazoo_web/src/kz_html_xss.erl
dhuynh-Kazoo: Aslo in the `kz_html_sanitizer:scrub/1` I already changed `[]` to `undefined`

kazoo_web/src/kz_html_xss.erl
dhuynh-Kazoo: Hi @jamesaimonetti , I am not sure I get you well. But the group can contain many pairs like `property: value`. For example, `color: red; text-align: center;`. So I think it is better to check the position of each pair. Let me know if I do not get you.

kazoo_web/src/kz_html_xss.erl
jamesaimonetti: In KAZOO, we put single quotes around all atoms, rather than relying on the implicit assumption that `true` is an atom, we make it explicit with `'true'`

kazoo_web/src/kz_html_xss.erl
jamesaimonetti: I wonder if it would make sense, at this point, to just iterate over things looking for an invalid item and return an error message with the bad value. You could even do a local `throw({error, Item})` to escape the recursion and catch it at the initial call to return.

Another option would be to fold over the items instead and add to the accumulator as you go.

The idea being to avoid iterating over the list again to flatten it.

kazoo_web/src/kz_html_xss.erl
jamesaimonetti: Sure, but compare:

```erlang
1> re:run(<<"color:red; text-align: center;">>,<<"([-\\w]+)\\s*:\\s*([^:;]*)">>,[global]).
{match,[[{0,9},{0,5},{6,3}],[{11,18},{11,10},{23,6}]]}

2> re:run(<<"color:red; text-align: center;">>,<<"([-\\w]+)\\s*:\\s*([^:;]*)">>,[global, {capture, all_but_first, binary}]).
{match,[[<<"color">>,<<"red">>],
        [<<"text-align">>,<<"center">>]]}
```
No need to do all the lookups to find the matches when `re` can do it for you.

kazoo_web/src/kz_html_xss.erl
dhuynh-Kazoo: Hi @jamesaimonetti, I think both ways are the same because in case we have SubNode in a Node, We need to call the origin function inside, so we have to add the found items to the accumulator and flatten to get all

kazoo_schemas/src/kz_json_schema_extensions.erl
```diff
@@ -157,6 +160,15 @@ extra_validation(_Key, _Value, State) ->
     lager:info("extra validation of ~s not handled for value ~p", [_Key, _Value]),
     State.
 
+announcement_validation(Value, State) ->
+    case kz_html_xss:has_xss_items(Value) of
+        false -> State;
```
jamesaimonetti: tickies

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,233 @@
...
+    HTML3 = re:replace(HTML2, "&lt;/html_sanitizer&gt;", "", [{'return', 'binary'}, 'global']),
+    after_to_html(HTML3);
+to_html(Token) -> to_html([Token]).
+
+remove_cdata_sections_before_scrub(HTML) ->
+    re:replace(HTML, "\\<\\!\\[CDATA\\[", "", ['global', {'return', 'binary'}]).
+
+before_parse(HTML) ->
+    HTML_1 = re:replace(
+               HTML,
```
jamesaimonetti: commas should prefix the arg:
```erlang
re:replace(HTML
          ,<<"...regex...">>
          ,[{'return', 'binary'}, 'global']
          )
```

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,233 @@
...
+    re:replace(
+      HTML2,
+      "(\&gt\;|>)(\ +)(" ++ ?REPLACEMENT_TAB ++ ")(\t+)(\&lt\;|<)",
+      "\\1\\4\\5",
+      [{'return', 'binary'}, 'global']
+     ).
+
+-spec traverse(list() | html_node() | tuple()) -> html_node() | html_attr().
+traverse(Nodes) when is_list(Nodes)->
+    [Res || Node <- Nodes, Res <- [handle_node(Node)], Res =/= 'undefined'];
```
jamesaimonetti: generators and filters should be on their own lines:

```erlang
    [Res || Node <- Nodes,
            Res <- [handle_node(Node)],
            Res =/= 'undefined'
    ];
```

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,233 @@
...
+    scrub({Tag, Attr, Children_1});
+handle_node({Token, Children}) ->
+    Children_1 = traverse(Children),
+    Children_2 = collapse_list(Children_1),
+    scrub({Token, Children_2});
+handle_node(Any) -> scrub(Any).
+
+scrub({<<"script">>, _, _}) -> 'undefined';
+scrub({<<"style">>, Attrs, Text}) ->
+    Fun = fun({Attr, Value}) ->
```
jamesaimonetti: In general, when a `fun()` doesn't enclose vars from the outer scope, I would prefer these to be named functions instead:

```erlang
is_style_attribute_valid({Attr, _Value}) ->
  lists:member(Attr, [...]).

scrub({<<"style">>, Attrs, Text}) ->
    Attrs1 = [Attr || Attr <- Attrs, is_style_attribute_value(Attr)],
    {<<"style">>, Attrs1, scrub_css(Text)};
```

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,233 @@
...
+                      'true' -> {Attr, Value};
+                      'false' -> 'undefined'
+                  end
+          end,
+    Attrs_1 = [X || X <- Attrs, Fun(X) =/= 'undefined'],
+    {<<"style">>, Attrs_1, scrub_css(Text)};
+scrub({Tag, Attrs, Data}) ->
+    Attrs_1 = [Res || X <- Attrs,
+                      Res <- [handle_attribute_and_value(X)],
+                      Res =/= 'undefined'],
```
jamesaimonetti: drop end of list comprehension to new line

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,233 @@
...
+    {<<"style">>, scrub_css(Value)};
+handle_attribute_and_value({Attribute, Value}) ->
+    case lists:member(Attribute, ?ATTRIBUTES_CAN_TRIGGER_XSS) of
+        'true' -> scrub_attribute_with_reference_src(Attribute, Value);
+        'false' -> {Attribute, Value}
+    end.
+
+scrub_attribute_with_reference_src(_Attr,<<"&", _/binary>>) -> [];
+scrub_attribute_with_reference_src(Attr, <<"data:", _/binary>> = Src) ->
+    case re:run(Src, ?ALLOWED_MIMETYPES_SOURCE, [global]) of
```
jamesaimonetti: If you run `make elvis` you should get warnings about all untickied atoms that KAZOO cares about.

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,233 @@
...
+    Text1 = re:replace(
+              Text,
+              "(\\/\\*|\\*\\/|<!--|-->)", " ",
+              [{'return', 'list'}, 'global']
+             ),
+
+    Fun = fun([SubMatched, Property, Value], Acc) ->
+                  Replace =
+                      case validate_css_keyword(Property) of
+                          'true' -> NewVal = scrub_css_val(Value),
```
jamesaimonetti: I don't think the temp var is necessary here, why not `[Property, ": ", scrub_css_val(Value)]`?

Also, multi-line clauses should start on new lines under the clause

kazoo_web/src/kz_html_sanitizer.erl
```diff
@@ -0,0 +1,233 @@
...
+-spec scrub_css(term()) -> term().
+scrub_css([]) -> [];
+scrub_css([Text]) ->
+    Text1 = re:replace(
+              Text,
+              "(\\/\\*|\\*\\/|<!--|-->)", " ",
+              [{'return', 'list'}, 'global']
+             ),
+
+    Fun = fun([SubMatched, Property, Value], Acc) ->
```
jamesaimonetti: This should be a named function

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,130 @@
...
+    find_xss_in_css(Val);
+find_xss_in_attribute({Attribute, Value}) ->
+    case lists:member(Attribute, ?ATTRIBUTES_CAN_TRIGGER_XSS) of
+        'true' -> find_xss_in_val(Attribute, Value);
+        'false' -> 'false'
+    end.
+
+find_xss_in_val(_Attr,<<"&", _/binary>>) -> 'true';
+find_xss_in_val(_Attr, <<"data:", _/binary>> = Src) ->
+    'nomatch' == re:run(Src, ?ALLOWED_MIMETYPES_SOURCE, ['global']);
```
jamesaimonetti: In general, KAZOO prefers to use `=:=` and `=/=` unless there's a specific need for `==` instead.

kazoo_web/src/kz_html_xss.erl
```diff
@@ -0,0 +1,130 @@
...
+           Text1,
+           "([-\\w]+)\\s*:\\s*([^:;]*)",
+           ['global', {'capture', 'all_but_first', 'binary'}]
+          ) of
+        {'match', List} -> lists:any(Fun, List);
+        'nomatch' -> 'false'
+    end.
+
+find_xss_property_val(Val) ->
+    Fun = fun([Matched]) ->
```
jamesaimonetti: named fun

PR #613, opened 2023-03-08
fix including header from other apps

icehess:


manushiperera:
This fix is already merged to master with #684 
But it is used `include_file` instead of `include_lib` in `kz_date` module.
Was there any specific reason to use `include_file`?

icehess:
@manushiperera Yes. Erlang idiom is too use `include` with relative path to the the source code itself, and `include_file` with relative path to header path (usually `include` dir) in an OTP app.

We hack this in our make files but adding any imaginable paths when calling compiler, but tooling rebar3 and erlang_ls expect the developer is following the common pattern and use these in appropriate ways.

This is fix in the other PR when it was introducing rebar as alternative build tools, and was fixing the compile issue there.

Closing this anyway as this is not already fixed.

PR #624, opened 2023-03-21
KZOO-308: added functionality to create call record from channel_create

nkarunarathne:
added kapps_call ability to create the call from channel_create so that that event can be used to start recording via kapps_call_recording


jamesaimonetti:
The Call record created from CHANNEL_CREATE won't have a call control queue available, so commands using kapps_call_command (for instance) won't work. I assume this is for use in ecallmgr only?

k-anderson:
@jamesaimonetti you are correct but the presedent already exists as we have "from_route_req", ect.  There are some places where having the Call record early is valid, because util functions use call parameters to calculate something.

Simple example from cf_route_req:

```
    Call = kapps_call:exec(Routines, kapps_call:from_route_req(RouteReq)),
    case is_binary(kapps_call:account_id(Call))
```

kazoo_call/src/kapps_call_recording.erl
```diff
@@ -122,29 +122,34 @@ record_call_command(Data, Call) ->
...
-            ,{<<"Media-Recording-ID">>, MediaDocId}
-            ,{<<"Record-Sample-Rate">>, SampleRate}
-            ,{<<"Record-Min-Sec">>, kz_term:to_binary(RecordMinSec)}
-            ,{<<"Media-Recorder">>, Recorder}
-            ,{<<"Recording-Variables">>, kz_json:from_list(Vars)}
-            ,{<<"Call-ID">>, CallId}
-            ,{<<"Msg-ID">>, kz_binary:rand_hex(16)}
-            | kz_api:default_headers(<<"call">>, <<"command">>, ?APP_NAME, ?APP_VERSION)
-            ],
+    Media = props:filter_undefined(
```
lazedo: no need for `props:filter_undefined` as `kz_json:from_list/1` already does it, any reason for this change ?

kazoo_call/src/kapps_call_recording.erl
```diff
@@ -122,29 +122,34 @@ record_call_command(Data, Call) ->
     RecordMinSec = kz_json:get_integer_value(<<"record_min_sec">>, Data, DefaultRecordMinSec),
     {Year, Month, _} = erlang:date(),
     CallId = kapps_call:call_id(Call),
-    RecordingId = kz_binary:rand_hex(16),
+    RecordingId = kz_json:get_ne_binary_value(<<"media_id">>, Data, kz_binary:rand_hex(16)),
```
lazedo: if the value configured doesn't change it will try create duplicate docs, why is this being changed ?

lazedo (changes requested):
media_id and props:filter_undefined should be changed

kazoo_call/src/kapps_call_recording.erl
```diff
@@ -126,12 +126,14 @@ record_call_command(Data, Call) ->
     MediaDocId = ?MATCH_MODB_PREFIX(kz_term:to_binary(Year), kz_date:pad_month(Month), RecordingId),
     DefaultMediaName = get_media_name(kz_binary:rand_hex(16), Format),
     MediaName = kz_json:get_value(?RECORDING_ID_KEY, Data, DefaultMediaName),
-    FollowTransfer = kapps_call:kvs_fetch('recording_follow_transfer', 'false', Call),
-    Recorder = media_recorder(Data, Call),
+    FollowTransfer = kz_json:is_true(<<"follow_transfer">>, Data, kapps_call:kvs_fetch('recording_follow_transfer', 'false', Call)),
+    Recorder = kz_json:get_ne_binary_value(<<"media_recorder">>, Data, media_recorder(Data, Call)),
+    RecordAsStereo = kz_json:is_true(<<"stereo">>, Data),
```
lazedo:
```suggestion
    RecordAsStereo = kz_json:get_boolean_value(<<"stereo">>, Data),
```

lazedo: the default is to record with stereo in the absence of `Channels-As-Stereo`, this is changing behavior

PR #645, opened 2023-04-06
KCRO-181: ignore empty mms body with at least a file

icehess:
This updates the schema becuase of relaxing MMS schema in KCRO-181.

cb_mms still checks properly that body is indeed none-empty.


icehess:
CI is failing because of circular pr dependency on https://github.com/2600hz/kazoo-crossbar/pull/463

PR #646, opened 2023-04-06
Add PropEr test to confirm `kz_json_generators:deep_object` input is valid

danielfinke:
- encode then decode generated object, expecting no `jiffy` decoding errors
- update `kz_json_generators:deep_object` to only produce terms that are or contain valid UTF-8 strings
- `kz_json:map/2` was discovered during `prop_utf8_encoding/0` to produce objects with duplicate keys, which are not valid, if the mapper fun returns the same key more than once. Those keys should be overwritten.
  - add PropEr test for this fix
- `flat_proplist()` actually has `path()` keys
  - `path()` can also be a mix of bins/ints, e.g. `kz_json:flatten_deep({[{<<"a">>, [1]}]}).`
- `kz_json:set_value1/4` crashed with `badarg` when trying to set a value in an array using a key that is not convertible into an int. This case is generated by PropEr when using the new `flat_proplist()` defn.
- remove unused `kz_json_generators:non_empty_object/0`


kazoo_stdlib/src/kz_json.erl
```diff
@@ -1689,16 +1696,15 @@ set_value(Keys, Value, JObj, Options) when is_list(Keys) ->
...
 
--spec set_value1(keys(), json_term() | 'null', object() | objects(), set_value_options()) -> object() | objects().
-set_value1([Key|_]=Keys, Value, [], Options) when not is_integer(Key) ->
+-spec set_value1(path(), json_term() | 'null', object() | objects(), set_value_options()) -> object() | objects().
+set_value1([Key|_]=Keys, Value, JObjs, Options) when is_list(JObjs), not is_integer(Key) ->
     set_value1(Keys, Value, new(), Options);
 set_value1([Key|T], Value, JObjs, Options) when is_list(JObjs) ->
-    Key1 = kz_term:to_integer(Key),
-    case Key1 > length(JObjs) of
+    case Key > length(JObjs) of
```
danielfinke, line 1707: This change needs some discussion before this is merged - it changes
```erlang
Opts = #{'deep' => 'true'},
kz_json:set_value([<<"a">>, <<"2">>], 0, kz_json:set_value([<<"a">>, 1], 1, kz_json:new(), Opts), Opts).
```
from
`{[{<<"a">>,[1,0]}]}` to `{[{<<"a">>,{[{<<"2">>,0}]}}]}`, meaning binary keys are not coerced into integer indexes.

Note that it fixes
```erlang
Opts = #{'deep' => 'true'},
kz_json:set_value([<<"a">>, <<"b">>], 0, kz_json:set_value([<<"a">>, 1], 1, kz_json:new(), Opts), Opts).
```
so it no longer throws a `badarg` error. I could add a special case to check if it's possible to coerce to integer, though.

jamesaimonetti (approved):
Initial review is positive for me

PR #658, opened 2023-04-20
Fix `kz_json_generators:utf8_json_string()`

danielfinke:
It was still possible for generated strings to be invalid UTF8 in really rare cases - `16#D800` to `16#DFFF` are invalid UTF8 codepoints, for example. `proper_unicode:utf8()` handles the edge cases.


jamesaimonetti (approved):
Nice!

PR #659, opened 2023-04-20
AXS-94: handle error during app fetch

icehess:


icehess:
close in favor of https://github.com/2600hz/kazoo-core/pull/699

PR #661, opened 2023-04-28
CEW-9: Handling missing parameter 'open_in_browser' in website API response

Navoda626:
CEW-9: Handling missing parameter 'open_in_browser' in website API response

Earlier, when fetching user websites through websits/listing_by_user view and when the website's account_wide:true, the parameter 'open_in_browser' is missed from the response.

Now, added missing open_in_browser parameter in websits/listing_by_user view in the case of account_wide=true. And the result will include that missing parameter.


nkarunarathne:
This looks good.

PR #663, opened 2023-05-04
KZOO-301: Add delete_bucket/2 fun to kazoo_token_buckets 

dhuynh-Kazoo:
This utility function will be called in cb_modules_utils for deleting the bucket to unlock account purpose


dhuynh-Kazoo:
This utility function will be called in cb_modules_utils for deleting the bucket to unlock account purpose

Main implementation for locking many failed login attempts: - https://github.com/2600hz/kazoo-crossbar/pull/467
Test case with user_auth login - https://github.com/2600hz/kazoo-properly/pull/150

kazoo_token_buckets/src/kz_token_bucket.erl
```diff
@@ -141,6 +143,14 @@ credit(Srv, Tokens) when is_integer(Tokens)
                          andalso Tokens > 0 ->
     gen_server:cast(Srv, {'credit', Tokens, get_callid()}).
 
+-spec delete_bucket(pid()) -> 'ok'.
+delete_bucket(Srv) ->
```
icehess: We already have `stop/1`, you can use that maybe? also can you add `get_callid()` to `stop` too so we get visibility in log for that request id?

icehess (changes requested):

kazoo_token_buckets/src/kz_buckets.erl
```diff
@@ -113,6 +113,18 @@ consume_tokens(<<_/binary>> = Key, Count, StartIfMissing) when is_integer(Count)
 consume_tokens(App, Key, Count, StartIfMissing) ->
     consume_tokens(App, Key, Count, StartIfMissing, fun kz_token_bucket:consume/2).
 
+-spec delete_bucket(kz_term:ne_binary(), kz_term:ne_binary()) -> boolean().
+delete_bucket(App, Key) ->
+    case get_bucket(App, Key) of
+        'undefined' -> 'false';
+        Srv ->
+            CallId = kz_token_bucket:get_callid(),
```
jamesaimonetti: This isn't necessary, the proc's callid is set higher up (in Crossbar or Callflow) so logging here will reflect that

PR #665, opened 2023-05-08
KZOO-379: Add a `kapps_maintenance` command to report inconsistencies in versions of deployed apps

barndon:
This PR adds a module to `kazoo_apps` called `kapps_versions` which contains functions for querying, analyzing, and reporting the versions of Kazoo apps deployed in a given environment.  This module is then used to implement two new `kapps_maintenance` commands:

- `show_kapps_versions`: prints the names and versions of all Kazoo apps on all nodes
- `check_kapps_versions`: prints warnings if multiple versions of a Kazoo app are found in the environment

### The Data Model
`kapps_versions` relies on a new record, `#node_app_version{}`, which holds the node, app name, app version, app description, and startup time.

Generally speaking, the functions in `kapps_versions` operate on lists and maps containing `#node_app_version{}` records.  A common type referenced in the code is a `group_map()` which is a map (potentially deeply nested) where the innermost value is a list of `#node_app_version{}`.  Group maps are intended to model hierarchies of categories.

`kapps_versions` contains functions for filtering, mapping, and folding lists and group maps.  All of these functions take plain functions as arguments, just like the `lists` and `maps` modules; many also support a shorthand notation where atoms can be used to indicate mapping and predicate functions built into `kapps_version`.

While most of the built-in mapping and predicate functions operate on `#node_app_version{}` records, it is important to note that the filtering and grouping functions of `kapps_versions` can be applied to _any_ type -- one need only supply the appropriate mapping and predicate functions.  The shorthand notation is easily extended to include new functions.

### How Version Information Is Gathered

Each node running `kazoo_nodes` broadcasts a heartbeat message via AMQP.  Historically, this message included only the names and startup times of Kazoo applications running on the node.  In this PR, the list of running apps and their startup times is merged with the list of Kazoo apps _loaded_ on the node to produce a list of `#node_app_version{}`.  This list is added to the heartbeat payload in a new field called `kapps_loaded`.

The `kapps_loaded` list is not guaranteed to contain a version string or a startup time for every app.  Here's why:

1. FreeSWITCH doesn't report its version or startup time. If such a `#node_app_version{}` is in the list, it means that FreeSWITCH is in the environment and was running recently.
2. Versions of Kazoo apps that predate this PR will not report their versions.  For these records, the version will be `<<"unknown">>`.
3. Kazoo apps that have been stopped or were never started may have a startup time of `undefined`.

#### List all the versions on all the nodes:
```erlang
L = kapps_versions:list().
```

#### Filter for a particular node, the hard way:
```erlang
F1 = fun(#node_app_version{node=N}) -> N =:= 'kazoo_apps@aio.craterhill.io' end.
kapps_versions:filter([F1], L).
```

#### Filter for a particular node, the easy way:
```erlang
kapps_versions:filter([{'node', <<"kazoo_apps@aio.craterhill.io">>}], L).
```
Notice that filters are supplied in a list.  The list is interpreted as a logical `and`.  For example, to get all the apps on a particular node where the version is less than 5.0:
```erlang
 kapps_versions:filter([{'node', <<"kazoo_apps@aio.craterhill.io">>}, {'version_lt', <<"5.0">>}], L).
```
#### Compound Filters

Logical expressions are supported with the predicates `and`, `or`, and `not`:
```erlang
kapps_versions:filter([{'or', [{'name', <<"webhooks">>}, {'name', <<"trunkstore">>}]}], L).
[#node_app_version{node = 'kazoo_apps@smflt006.craterhill.io',
                   name = <<"webhooks">>,
                   description = <<"Webhooks: HTTP events from KAZOO to you">>,
                   version = <<"4.0.0">>,startup = 63852778295},
 #node_app_version{node = 'kazoo_apps@smflt006.craterhill.io', 
                   name = <<"trunkstore">>,
                   description = <<"Trunkstore: Authentication and SIP Routing for PBXes">>,
                   version = <<"4.0.0">>,startup = 63852778295},
 #node_app_version{node = 'kazoo_apps@aio.craterhill.io',
                   name = <<"trunkstore">>,description = <<"unknown">>,
                   version = <<"unknown">>,startup = 63852169274},
 #node_app_version{node = 'kazoo_apps@aio.craterhill.io',
                   name = <<"webhooks">>,description = <<"unknown">>,
                   version = <<"unknown">>,startup = 63852169274}]
```

#### Organize `L` into groups by node name, the hard way:
```erlang
M1 = fun(#node_app_version{node=N}) -> N end.
kapps_versions:groups_by([M1], L).
```
The result is a map where the keys are nodes and the values are lists of `#node_app_version{}`, for example:
```erlang
#{'ecallmgr@aio.craterhill.io' =>
      [#node_app_version{node = 'ecallmgr@aio.craterhill.io',
                         name = <<"ecallmgr_extension">>,description = <<"unknown">>,
                         version = <<"unknown">>,startup = 63852169262},
       #node_app_version{node = 'ecallmgr@aio.craterhill.io',
                         name = <<"ecallmgr">>,description = <<"unknown">>,
                         version = <<"unknown">>,startup = 63852169261}],
  'freeswitch@aio.craterhill.io' =>
      [#node_app_version{node = 'freeswitch@aio.craterhill.io',
                         name = <<"freeswitch">>,description = <<"unknown">>,
                         version = <<"unknown">>,startup = undefined}],
  'kamailio@aio.craterhill.io' =>
      [#node_app_version{node = 'kamailio@aio.craterhill.io',
                         name = <<"kamailio">>,description = <<"unknown">>,
                         version = <<"unknown">>,startup = 63852169246}],
  'kazoo_apps@aio.craterhill.io' =>
      [#node_app_version{node = 'kazoo_apps@aio.craterhill.io',
                         name = <<"webhooks">>,description = <<"unknown">>,
                         version = <<"unknown">>,startup = 63852169274},
       ...
      ],
  ...
}
```
#### Grouping by node, the easy way:
```erlang
kapps_versions:groups_by(['node'], L).
```
Group maps can be arbitrarily deep.  Given 1 mapping function, the output is a map of lists.  Given 2 mapping functions, the output is a map of maps of lists.  Given N mapping functions, the output is a map of maps N keys deep, and the value is a list.  For example, grouping by node and version (the easy way):
```erlang
kapps_versions:groups_by(['node', 'version'], L).
```

The result looks like this:
```erlang
#{'ecallmgr@aio.craterhill.io' =>
      #{<<"unknown">> =>
            [#node_app_version{node = 'ecallmgr@aio.craterhill.io',
                               name = <<"ecallmgr_extension">>,description = <<"unknown">>,
                               version = <<"unknown">>,startup = 63852169262},
             #node_app_version{node = 'ecallmgr@aio.craterhill.io',
                               name = <<"ecallmgr">>,description = <<"unknown">>,
                               version = <<"unknown">>,startup = 63852169261}]},
  'freeswitch@aio.craterhill.io' =>
      #{<<"unknown">> =>
            [#node_app_version{node = 'freeswitch@aio.craterhill.io',
                               name = <<"freeswitch">>,description = <<"unknown">>,
                               version = <<"unknown">>,startup = undefined}]},
  'kamailio@aio.craterhill.io' =>
      #{<<"unknown">> =>
            [#node_app_version{node = 'kamailio@aio.craterhill.io',
                               name = <<"kamailio">>,description = <<"unknown">>,
                               version = <<"unknown">>,startup = 63852169246}]},
  'kazoo_apps@aio.craterhill.io' =>
      #{<<"unknown">> =>
            [#node_app_version{node = 'kazoo_apps@aio.craterhill.io',
                               name = <<"webhooks">>,description = <<"unknown">>,
                               version = <<"unknown">>,startup = 63852169274},
         ...
      ]
   }
...
]
```
**Order matters!**  `groups_by(['node', 'version'], L)` is not the same as `groups_by(['version', 'node'], L)`.

#### Filtering Groups

Group maps can also be filtered using the `filter_groups/2` function.  This function takes a list of predicates and a group map and produces a group map containing only the groups for which all the predicates are true.  For example, to find all of the nodes where both Webhooks and Trunkstore are running:
```erlang
L2 = kapps_versions:filter([{'names', [<<"webhooks">>, <<"trunkstore">>]}], L).
G1 = kapps_versions:groups_by(['node'], L2]).
kapps_versions:filter_groups([{'count', 2}], G1).
```
This same result can be achieved more succinctly using `groups_by/3`:

```erlang
L2 = kapps_versions:filter([{'names', [<<"webhooks">>, <<"trunkstore">>]}], L).
kapps_versions:groups_by(['node'], L2, [{'count', 2}]).
```


kazoo_amqp/src/api/kapi_nodes.erl
```diff
@@ -64,6 +64,7 @@ advertise_definition() ->
                                                             ,<<"Used-Memory">>
                                                             ,<<"Version">>
                                                             ,<<"WhApps">>
+                                                            ,<<"KappsLoaded">>
```
jamesaimonetti: Nit: `Kapps-Loaded` instead?

kazoo_nodes/src/kz_nodes_listener.erl
```diff
@@ -371,6 +373,7 @@ from_json(JObj, State) ->
             ,md5=kz_json:get_value(<<"md5">>, JObj)
             ,expires=kz_term:to_integer(kz_json:get_integer_value(<<"Expires">>, JObj, 0) * ?FUDGE_FACTOR)
             ,kapps=kapps_from_json(kz_json:get_value(<<"WhApps">>, JObj, []))
+            ,kapps_loaded=kapps_version_info_from_json(kz_json:get_value(<<"KappsLoaded">>, JObj, []))
```
jamesaimonetti: when possible, use typed getters like `kz_json:get_list_value/3` here

kazoo_nodes/src/kz_nodes_listener.erl
```diff
@@ -432,6 +435,28 @@ whapp_info_to_json(#whapp_info{startup=Start, roles=Roles}) ->
       ,{<<"Roles">>, Roles}
       ]).
 
+-spec kapps_version_info_from_json(kz_term:api_terms()) -> kz_types:kapps_version_info().
```
jamesaimonetti: Probably `kz_json:objects()` would be a more appropriate type. `api_terms()` typically refers to payloads meant for AMQP (which can be proplists or JObjs)

kazoo_nodes/src/kz_nodes_listener.erl
```diff
@@ -432,6 +435,28 @@ whapp_info_to_json(#whapp_info{startup=Start, roles=Roles}) ->
...
+kapps_version_info_from_json(Versions) when is_list(Versions) ->
+    [kapps_version_from_json(Version) || Version <- Versions].
+
+-spec kapps_version_from_json(kz_term:api_terms()) -> kz_types:kapps_version_info().
+kapps_version_from_json(JVer) ->
+    #kapps_version{
+       name=kz_json:get_binary_value(<<"Name">>, JVer)
```
jamesaimonetti: Can either name or version be the empty binary? `get_ne_binary_value` (and the type `kz_term:ne_binary()` would indicate better that `<<>>` is not valid.

kazoo_nodes/src/kz_nodes_listener.erl
```diff
@@ -432,6 +435,28 @@ whapp_info_to_json(#whapp_info{startup=Start, roles=Roles}) ->
...
+      }.
+
+-spec kapps_version_info_to_json(kz_types:kapps_version_info()) -> kz_json:object().
+kapps_version_info_to_json(KappsVersions) ->
+    [kapps_version_to_json(KappsVer) || KappsVer <- KappsVersions].
+
+-spec kapps_version_to_json(kz_types:kapps_version()) -> kz_json:object().
+kapps_version_to_json(#kapps_version{name=Name, version=Version}) ->
+    kz_json:from_list(
+      [{<<"Name">>, kz_term:to_binary(Name)},
```
jamesaimonetti: These values should already be binaries so I wouldn't bother with casting here - but its also effectively a no-op so no biggie.

kazoo_stdlib/include/kz_records.hrl
```diff
@@ -4,10 +4,15 @@
                     ,roles = [] :: kz_term:ne_binaries()
                     }).
 
+-record(kapps_version, {name :: kz_term:api_binary()
```
jamesaimonetti: `api_ne_binary()` if name/version cannot be `<<>>`

jamesaimonetti (changes requested):

kazoo_stdlib/src/kz_application.erl
```diff
@@ -131,6 +133,21 @@ running_apps() ->
         is_kapp(App)
     ].
 
+-spec loaded_apps() -> [kapp_version()].
+loaded_apps() ->
+    [
+     kapp_version(AppData)
+     || {Name, _, _} = AppData <- application:loaded_applications(),
```
icehess: Just bringing our conversion from slack message here:

> https://www.erlang.org/doc/reference_manual/code_loading.html#code-loading
>
> so in CI we don't have all apps obviously so bootscript is minimal.
> we install app on servers individually via package manager.
> and we allow anyone to make a choice to start the app or not
> so by that doc definition if we don't reference the app (either in bootscript or try to load or start the app) it is not loaded

jamesaimonetti: How about instead, using `code:get_path()` to list available apps on the filesystem that the VM knows about?

jamesaimonetti: for instance https://github.com/2600hz/kazoo-core/pull/665/commits/9a7223fd70bc831aef73ff4cee074f8b5710881a

kazoo_nodes/src/kz_nodes_listener.erl
```diff
@@ -371,6 +378,7 @@ from_json(JObj, State) ->
             ,md5=kz_json:get_value(<<"md5">>, JObj)
             ,expires=kz_term:to_integer(kz_json:get_integer_value(<<"Expires">>, JObj, 0) * ?FUDGE_FACTOR)
             ,kapps=kapps_from_json(kz_json:get_value(<<"WhApps">>, JObj, []))
+            ,kapps_loaded=kapp_versions_from_json(kz_json:get_value(<<"Kapps-Loaded">>, JObj, []))
```
jamesaimonetti: This defaults to empty list if key isn't found, but kapp_versions_from_json says the value should be a JSON object (not a list of objects). Clarify which?

kazoo_stdlib/src/kz_application.erl
icehess: This a start, but still `code:get_path/1` returns whatever path is _loaded_ into code server. If private app is not referenced anywhere then code:get_path still have no knowledge of it.

The better approach is to scan the the kazoo lib folder i guess.

Still i don't believe this issue has impact for appex as private app is not loaded not not physically present anyway

icehess: Actually never mind, `code:get_path/1` is already scanning the all erl lib folders on vm startup.

PR #667, opened 2023-05-15
Core changes related to parallel properly work

jamesaimonetti:
- Types updates
- Addressed infinite loop when ensuring an account db is deleted
- added system flag to toggle DNS validation during JSON schema validation
- Removed anon fun and calculated just milliseconds elapsed without needing floats


jamesaimonetti:
schema changes: https://github.com/2600hz/kazoo-crossbar/pull/476

PR #668, opened 2023-05-15
Use erlang.mk to build SUP

jamesaimonetti:


jamesaimonetti:
Wants https://github.com/2600hz/kazoo5/pull/108

PR #673, opened 2023-05-30
KZOO-383: Use paginated reads to remove orphaned services

barndon:
This PR introduces `kz_datamgr:fold_all_docs/4` and `/5`.  These functions call `lists:foldl/3` over blocks of documents from a given database until all documents are processed.  The accumulator is preserved between blocks and the final accumulator is returned.

`kazoo_services_maintenance:remove_orphaned_services/0` is updated to use `kz_datamgr:fold_all_docs/4`.


icehess:
@barndon  Just fyi, have you seen ~`kz_view:find`~ (sorry `kz_view:find_batch`, this one returns cursor and you can fold over it and call `find_next`) and its kz_view_cursor map/fold/foreach functions? You can check crossbar_view too since it uses kz_view heavily (wlp not that map/fold thing but in general crossbar uses kz_view).

jamesaimonetti:
Curious why fold_all_docs didn't use paginate_results for the fetching and tracking of next start key.

See for instance https://github.com/2600hz/kazoo-tasks/blob/master/src/modules/kt_modb_creation.erl#L96

kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -990,12 +990,20 @@ prepare_service_doc(JObj) ->
 
 -spec remove_orphaned_services() -> 'no_return'.
 remove_orphaned_services() ->
-    {'ok', ServiceDocs} = kz_datamgr:all_docs(?KZ_SERVICES_DB),
-    Count = lists:foldl(fun maybe_remove_orphan/2, 0, ServiceDocs),
+    BatchSize = kz_datamgr:max_bulk_read(),
+    {'ok', Cursor} = kz_view:find_batch(?KZ_SERVICES_DB, 'all_docs', [], [{'batch_size', BatchSize}]),
```
icehess, line 994: or you can just simply use `kz_view_cursor:mapfold/4` to get the count and report. If you search  for `kz_view_cursor:foreach` in core or application you will find some samples on how easy and handy are those `kz_view_cursor` functions.

icehess (changes requested):

PR #675, opened 2023-06-13
PS-35: create call forward endpoint when kz_endpoint:build/3 is used

jamesaimonetti:


jamesaimonetti:
updating docs to reflect configs required to make this work

PR #677, opened 2023-06-15
AXS-87: update bluepak and remove appex_cb_router

icehess:
The task of setting up the UI apps and integrating to the crossbar for serving their assets file has been moved to the package themselves.

This cleans up appex_cb_router code from kazoo_appex.

BluePAK packager now uses templates to generate the UI app wrappers which includes the app specific crossbar integration. Also BluePAK now sanity checks `.app` files and packages to make sure they have application callback modules.


kazoo_appex/src/appex_bluepak.erl
```diff
@@ -175,8 +160,9 @@ adjust_dot_app(#load_app{app_spec={'application', App, AppConfig}
                         }=LoadApp) ->
     {AppMod, StartArgs} = props:get_value('mod', AppConfig),
     AppEnv = props:get_value('env', AppConfig, []),
-    NewEnv = props:set_values([{'app_mod', AppMod}
-                              ,{'parent_app', ParentApp}
+    NewEnv = props:set_values([{'kappex_app_mod', AppMod}
+                              ,{'kappex_parent_app', ParentApp}
+                              ,{'is_kappex', 'true'}
```
icehess: changed in https://github.com/2600hz/kazoo-core/pull/702/files 
```suggestion
                              ,{'from_kappex', 'true'}
```

PR #678, opened 2023-06-20
KZOO-397: Treat classifiers as an array (order preserving)

jamesaimonetti:
Versus an object which does not preserve order of keys when (de)serializing.


nkarunarathne:
overall code looks good to me.

PR #679, opened 2023-06-21
CCOW-41-Favorite-status-is-not-being-displayed-in-conferences-master

kdamas93:
Currently, when someone wants to get the users who have bookmarked a conference as a favorite is not possible, Although, in the conference document, the array of user ids is saved, it is not shown because there is no definition or default value in account-crossbar-listings.json.
The endpoint cb_conferences uses the view **<<"crossbar_listings/by_type_id">>** which does not show this field, due to this the proposed change is the content in this PR


jamesaimonetti:
Talked with @kdamas93 and this will go in a different direction, closing.

PR #685, opened 2023-07-11
[master] kcro-188-ListsApiEnhancements

kdamas93:
Related changes for lists API support, more detail can be found in this reference
https://github.com/2600hz/kazoo-crossbar/pull/500


kazoo_documents/src/kzd_list_entries.erl
```diff
@@ -1,25 +1,18 @@
 %%%-----------------------------------------------------------------------------
 %%% @copyright (C) 2010-2023, 2600Hz
-%%% @doc
-%%% This Source Code Form is subject to the terms of the Mozilla Public
```
jamesaimonetti: Don't remove this

kazoo_documents/src/kzd_lists.erl
```diff
@@ -1,10 +1,6 @@
 %%%-----------------------------------------------------------------------------
 %%% @copyright (C) 2010-2023, 2600Hz
-%%% @doc
-%%% This Source Code Form is subject to the terms of the Mozilla Public
```
jamesaimonetti: Don't remove this

jamesaimonetti (changes requested):

kazoo_documents/src/kzd_list_entries.erl
kdamas93: This change has been reverted so that it has the original document.

kazoo_documents/src/kzd_lists.erl
kdamas93: This change has been reverted so that it has the original document.

kazoo_documents/src/kzd_list_entries.erl.src
```diff
@@ -37,65 +26,41 @@
...
-set_capture_group_length(Doc, CaptureGroupLength) ->
-    kz_json:set_value([<<"capture_group_length">>], CaptureGroupLength, Doc).
+-spec set_description(doc(), binary()) -> doc().
+set_description(Doc, Description) ->
+    kz_json:set_value([<<"description">>], Description, Doc).
 
--spec displayname(doc()) -> kz_term:api_ne_binary().
-displayname(Doc) ->
-    displayname(Doc, 'undefined').
+-spec history(doc()) -> kz_json:objects().
```
nkarunarathne: [] is not a json object. may need to change the return type in the spec.

maybe something like kz_json:objects() | []

kdamas93: change applied!

icehess (changes requested):
Do you also need to update `lists` views? if the `number` field is changed the view needs to be updated (if you change `lists` also update `crossbar-listings` too).

Also if `number` field is changed you may also need to search all apps and update the code there, for example this shows callflow needs to be update:

```
$ ag lists/match core/ applications/
....
....
applications/callflow/src/module/cf_cidlistmatch.erl
58:    case kz_datamgr:get_results(AccountDb, <<"lists/match_prefix_in_list">>, [{'keys', Keys}]) of

applications/callflow/src/module/cf_lookupcidname.erl
76:    case kz_datamgr:get_results(AccountDb , <<"lists/match_prefix_in_list">>, ViewOptions) of
```

Repaeat this for all other lists views.

kazoo_documents/src/kzd_list_entries.erl
```diff
@@ -33,65 +30,41 @@
 new() ->
     kz_json_schema:default_object(?SCHEMA).
 
--spec capture_group_key(doc()) -> kz_term:api_binary().
-capture_group_key(Doc) ->
-    capture_group_key(Doc, 'undefined').
+-spec contacts(doc()) -> kz_json:objects().
```
nkarunarathne, line 33: change the spec the way you done in history function.

return type ->  `kz_json:objects() | [].`

PR #691, opened 2023-07-20
crm shared module

kusalfernando:
This include CRM shared module for authentication and get caller info which is used by qubicle.


kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,64 @@
...
+%%%
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(kapi_crm).
+
+-include_lib("kz_amqp_util.hrl").
+-include("kapi_crm.hrl").
+
+%% CRM functions
+-export([authenticate/1, caller_info/1]).
```
jamesaimonetti: Please format according to project standards. Consult other modules for guidance if needed or read the [contributing guide](https://github.com/2600hz/kazoo5/blob/master/CONTRIBUTING.md)

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,64 @@
...
+authenticate(Args) ->
+    kz_amqp_worker:cast(Args, fun publish_authenticate/1).
+
+-spec publish_authenticate(kz_term:api_terms()) -> 'ok'.
+publish_authenticate(Args) ->
+    {'ok', Payload} = kz_api:build_message(Args, ?AUTH_HEADERS, []),
+    kz_amqp_util:basic_publish(?CRM_EXCHANGE, route(Args), Payload, ?DEFAULT_CONTENT_TYPE).
+
+-spec authenticate_resp(kz_json:object()) -> boolean().
+authenticate_resp(Message) ->
```
jamesaimonetti: Function names to return booleans should indicate so. `is_authentication_reply` for instance

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,64 @@
...
+%%% @copyright (C) 2023, 2600Hz
+%%% @doc
+%%% @author Kusal Fernando
+%%% This Source Code Form is subject to the terms of the Mozilla Public
+%%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%%
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(kapi_crm).
```
jamesaimonetti, line 11: Please consult other kapi modules, define the AMQP payload definitions, builders, validators, and publishers.

jamesaimonetti: The caller10 guide has a section all about kapi modules:
https://github.com/2600hz/docs-training/blob/666341a35bc525a16cfd2425225d3c91667ea818/caller10/06_callflows.org#the-kapi-module

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,263 @@
...
+-export([bind_q/2
+        ,unbind_q/2
+        ,declare_exchanges/0]).
+
+-export([api_definitions/0, api_definition/1]).
+
+-export([authenticate/1
+        ,authenticate_v/1
+        ,publish_authenticate/1
+        ,publish_authenticate/2]).
```
jamesaimonetti: Drop `]).` to newline as per project standards

kazoo_amqp/src/api/kapi_crm.hrl
```diff
@@ -0,0 +1,43 @@
...
+
+-define(CALLER_INFO_HEADERS, [<<"Caller-ID">>
+                             ,<<"Client-Id">>
+                             ,<<"Client-Secret">>
+                             ,<<"Instance-Url">>
+                             ,<<"Access-Token">>
+                             ]).
+
+-define(CALLER_INFO_RESPONSE_HEADERS, [<<"Id">>
+                                      ,<<"Caller_Name__c">>
```
jamesaimonetti: where is this `__c` suffix coming from?

jamesaimonetti (changes requested):

kazoo_amqp/src/api/kapi_crm.hrl
kusalfernando: It's coming from Salesforce and we don't format it yet as we are not sure about the format of the data yet.

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,268 @@
...
+              ,{fun kapi_definition:set_values/2
+               ,kapi_definition:event_type_headers(Category, EventName)
+               }
+              ,{fun kapi_definition:set_types/2, []}
+              ],
+    kapi_definition:setters(Setters).
+
+-spec authenticate_resp_definition() -> kapi_definition:api().
+authenticate_resp_definition() ->
+    EventName = <<"authenticate-reply">>,
```
jamesaimonetti: using underscores is consistent with other EventName settings

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,268 @@
...
+%%------------------------------------------------------------------------------
+%% @doc Get all API definitions of this module.
+%% @end
+%%------------------------------------------------------------------------------
+-spec api_definitions() -> kapi_definition:apis().
+api_definitions() ->
+    [authenticate_definition()
+    ,authenticate_resp_definition()
+    ,caller_info_definition()
+    ,caller_info_resp_definition()].
```
jamesaimonetti: Drop `]` to newline

jamesaimonetti (changes requested):

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,269 @@
...
+caller_info_resp_definition() ->
+    EventName = <<"caller_info_reply">>,
+    Category = <<"crm-salesforce">>,
+    Setters = [{fun kapi_definition:set_name/2, EventName}
+              ,{fun kapi_definition:set_friendly_name/2, <<"Salesforce Query Response">>}
+              ,{fun kapi_definition:set_description/2, <<"Caller info query response">>}
+              ,{fun kapi_definition:set_category/2, Category}
+              ,{fun kapi_definition:set_build_fun/2, fun caller_info_resp/1}
+              ,{fun kapi_definition:set_validate_fun/2, fun caller_info_resp_v/1}
+              ,{fun kapi_definition:set_required_headers/2, ?CALLER_INFO_RESPONSE_HEADERS}
```
icehess: I advise to not use macro here since you are not even share these headers anywhere else.

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,269 @@
...
+               ,kapi_definition:event_type_headers(Category, EventName)
+               }
+              ,{fun kapi_definition:set_types/2, []}
+              ],
+    kapi_definition:setters(Setters).
+
+-spec caller_info_resp_definition() -> kapi_definition:api().
+caller_info_resp_definition() ->
+    EventName = <<"caller_info_reply">>,
+    Category = <<"crm-salesforce">>,
```
icehess: hmm, use underscore to be consistence with other api

kazoo_amqp/src/api/kapi_crm.erl
icehess: do the same for other definition functions.

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,269 @@
...
+              ,{fun kapi_definition:set_category/2, Category}
+              ,{fun kapi_definition:set_build_fun/2, fun authenticate/1}
+              ,{fun kapi_definition:set_validate_fun/2, fun authenticate_v/1}
+              ,{fun kapi_definition:set_publish_fun/2, fun publish_authenticate/2}
+              ,{fun kapi_definition:set_required_headers/2, ?AUTH_HEADERS}
+              ,{fun kapi_definition:set_optional_headers/2, []}
+              ,{fun kapi_definition:set_values/2
+               ,kapi_definition:event_type_headers(Category, EventName)
+               }
+              ,{fun kapi_definition:set_types/2, []}
```
icehess: I suggest to always type your data. look at other kapi modules, like here you have `Client-ID` (pay attention to capital `D` here too). you can type it like:
```erlang
{<<"Client-ID">>, fun kz_term:is_binary/1} %% or kz_term:is_ne_binary/1 if you want to force be non empty string
```

icehess: also do the same for all other api definition functions.

kazoo_amqp/src/api/kapi_crm.hrl
```diff
@@ -0,0 +1,43 @@
...
+-define(CRM_ROUTE(A, E), <<"crm.", A/binary, ".", E/binary>>).
+
+-define(AUTH_HEADERS, [<<"Client-Id">>
+                      ,<<"Client-Secret">>
+                      ,<<"Instance-Url">>
+                      ]).
+
+-define(AUTH_RESPONSE_HEADERS, [<<"access_token">>]).
+
+-define(AUTH_RESPONSE_HEADERS_OPT, [<<"instance_url">>
```
icehess: Top level keys in all Kapi modules are CamelCase and use `-` , please do the same here and be consistence with the rest of kapi modules.

If you are just want to send the exact data you received from and external service, maybe you can just put those data inside `CRM-Response` with type `kz_json:is_object/1` in this case? What do you think?

This applies to the other headers here, apply this to all of them.

kazoo_amqp/src/api/kapi_crm.hrl
icehess: Look at my other comment, if you want to reply unchanged response from Salesforce then don't set those in top level, set them in `Salesforce-Response` as is `kz_json:is_object` type.

icehess (changes requested):

kazoo_amqp/src/api/kapi_crm.hrl
kusalfernando: I agree with that and I will go with `Auth-Response` and type is `kz_json:is_object/1`.

kazoo_amqp/priv/couchdb/schemas/kapi.crm.authenticate.json
```diff
@@ -0,0 +1,36 @@
...
+            "minLength": 1,
+            "type": "string"
+        },
+        "Client-Secret": {
+            "minLength": 1,
+            "type": "string"
+        },
+        "Event-Category": {
+            "enum": [
+                "crm-salesforce"
```
jamesaimonetti: I would have assumed event category is `crm` for all of these, and the particular CRM module (salesforce) does salesforcey stuff with the payload

kazoo_amqp/priv/couchdb/schemas/kapi.crm.caller_info.json
```diff
@@ -0,0 +1,45 @@
...
+            "minLength": 1,
+            "type": "string"
+        },
+        "Client-Secret": {
+            "minLength": 1,
+            "type": "string"
+        },
+        "Event-Category": {
+            "enum": [
+                "crm-salesforce"
```
jamesaimonetti: same here

kazoo_amqp/priv/couchdb/schemas/kapi.crm.caller_info_reply.json
```diff
@@ -0,0 +1,27 @@
...
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "_id": "kapi.crm.caller_info_reply",
+    "description": "AMQP API for crm.caller_info_reply",
+    "properties": {
+        "Caller-Info-Response": {
+            "type": "object"
+        },
+        "Event-Category": {
+            "enum": [
+                "crm-salesforce"
```
jamesaimonetti: here too

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,291 @@
...
+    authenticate_resp_definition();
+api_definition(<<"caller_info">>) ->
+    caller_info_definition();
+api_definition(<<"caller_info_resp">>) ->
+    caller_info_resp_definition().
+
+-spec authenticate_definition() -> kapi_definition:api().
+authenticate_definition() ->
+    EventName = <<"authenticate">>,
+    Category = <<"crm_salesforce">>,
```
jamesaimonetti: just `crm`

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,291 @@
...
+              ],
+    kapi_definition:setters(Setters).
+
+-spec caller_info_definition() -> kapi_definition:api().
+caller_info_definition() ->
+    EventName = <<"caller_info">>,
+    Category = <<"crm_salesforce">>,
+    Setters = [{fun kapi_definition:set_name/2, EventName}
+              ,{fun kapi_definition:set_friendly_name/2, <<"Query caller info">>}
+              ,{fun kapi_definition:set_description/2, <<"Query caller info in Salesforce">>}
```
jamesaimonetti, line 146: just `in a CRM` - remove salesforce references

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -0,0 +1,291 @@
...
+              ,{fun kapi_definition:set_category/2, Category}
+              ,{fun kapi_definition:set_build_fun/2, fun caller_info/1}
+              ,{fun kapi_definition:set_validate_fun/2, fun caller_info_v/1}
+              ,{fun kapi_definition:set_publish_fun/2, fun publish_caller_info/2}
+              ,{fun kapi_definition:set_required_headers/2, [<<"Caller-ID">>
```
jamesaimonetti, line 151: I wonder how CRM info like this is stored? If in the account database, I would suggest passing account ID / doc ID instead of sensitive client id/secret/token info. Let the CRM app fetch the sensitive stuff from the DB (or wherever)

jamesaimonetti (changes requested):
No salesforce references

kazoo_amqp/src/api/kapi_crm.erl
kusalfernando: This was planned with @mark2600 to pass the credentials into CRM when the request is made from qubicle.

mark2600: it is ultimately stored in the db, so this is certainly an optimization to be done in the future.  unless you have major concerns with this i would say we should move ahead as is for now, given the impending code freeze.

jamesaimonetti: Seems simple enough to change though. Instead of

```erlang
handle_crm(JObj, _Props) ->
    ClientId = kz_json:get_ne_binary_value(<<"Client-ID">>, JObj),
    ...
```

it becomes
```erlang
handle_crm(JObj, _Props) ->
    AccountId = kz_json:get_ne_binary_value(<<"Account-ID">>, JObj),
    CRMConfigId = kz_json:get_ne_binary_value(<<"Config-ID">>, JObj),
    {'ok', Config} = kz_datamgr:open_cache_doc(AccountId, CRMConfigId),

    ClientId = kzd_crm:client_id(Config),
    ....
```

PR #693, opened 2023-07-25
KCRO-190: Additional APIs to consume pvt_chat.getstream in user doc

Navoda626:
Created a new view "list_by_getstream" which will fetch the users whose pvt_chat.getstream.enabled == 'true' to facilitate the following crossbar API

**GET accounts/{ACCOUNT_ID}/users/getstream** <-- list all user documents with pvt_chat.getstream.enabled == true


nkarunarathne:
Looks good to me

kazoo_apps/priv/couchdb/account/users.json
```diff
@@ -76,6 +76,46 @@
                 "}"
             ]
         },
+        "list_by_getstream": {
```
jamesaimonetti, line 79: My initial thought here is - this view isn't necessary. Fetch users in crossbar with include_docs, then filter based on pvt_chat.getstream, and return the users having getstream configured.

Thoughts?

Navoda626: I used a view since user fetch will have a `features` parameter in user body, so thought its easy to use a view to generate that features parameter.

If we omit the features parameter what is better, using a view or filter after fetching all user documents?

Navoda626: Using an erlang filter after fetching using crossbar_listing

Navoda626:
Using an erlang filter after fetching using crossbar_listing. So this view is not required

PR #698, opened 2023-08-08
KCRO-191: Fix local available numbers listing

harenson:
Properly tests: https://github.com/2600hz/kazoo-properly/pull/176

When listing available local numbers via NPANXX query, the local carrier is not honoring the given NPANXX prefix, it is only honoring the first 3 digits (NPA). `knm_local`'s changes fix it.

Also:
- Add some documentation to `props:set_values/2` function.

**NOTE**: Only knm_local carrier module is failing when NPANXX prefix is given. Other carriers are working as expected.


kazoo_numbers/src/knm_search.erl
```diff
@@ -259,7 +259,7 @@ first(Options) ->
                ,{'offset', 0}
                ,{'normalized_prefix', normalized_prefix(Options)}
                ],
-    Opts = props:set_values(Defaults, Options),
+    Opts = props:set_values(Options, Defaults),
```
lazedo: undo this, its working as expected.

kazoo_numbers/src/knm_search.erl
```diff
@@ -336,6 +336,18 @@ next(Options) ->
...
+    %% prefix.
+    %%
+    %% Example of MatchSpec implementation in OTP@26:
+    %%  Prefix = normalized_prefix(Options),
+    %%  PrefixLen = size(Prefix),
+    %%  MatchSpec =
+    %%      ets:fun2ms(fun({Qid, {N, _ModuleName, _State, _CarrierData}} = Num)
+    %%          when Qid = QID
+    %%              andalso binary_part(N, 0, PrefixLen) =:= Prefix -> Num
+    %%      end).
```
lazedo: add the comment when we get to otp26, remove it for now

kazoo_numbers/src/carriers/knm_local.erl
```diff
@@ -91,8 +91,8 @@ do_find_numbers(<<"+",_/binary>>=Prefix, Quantity, Offset, AccountId, QID)
                     lager:debug("found available local numbers for account ~s in ~s"
                                ,[AccountId, DB]
                                ),
-                    Numbers = format_numbers(QID, JObjs),
-                    find_more(Prefix, Quantity, Offset, AccountId, length(Numbers), QID, Numbers);
+                    Numbers = format_numbers(QID, JObjs, Prefix),
+                    find_more(Prefix, Quantity, Offset, AccountId, length(JObjs), QID, Numbers);
```
lazedo, line 95: why the change ?

lazedo (changes requested):
defaults are working as expected, they override whatever the caller may have set on those params

kazoo_numbers/src/carriers/knm_local.erl
harenson: To move the _offset_. Avoids looping forever over the same non-matching numbers.

PR #699, opened 2023-08-14
AXS-101: bind for node heartbeats and adjust runnings apps accordingly

jamesaimonetti:
Preconditions:
 1. Appex client is configured to run
 2. `sup kapps_config appex_client is_true is_aio_cluster` returns true

Clusters using the app marketplace will, by default, run as all-in-one KAZOO servers - meaning any app running on node N will also run on node N+1.

When node heartbeats arrive from remote nodes, kappex_controller will compare their list of running kapps to the local nodes; if any apps are running remotely that are not running locally, kappex_controller will check if the app is loadable.

If loadable, the app is attempted to be started via normal kapps_controller interactions (spawned, to avoid slow-starting apps blocking kappex_controller).

If not loadable, currently, logging will take place indicating the issue but otherwise no action is taken.


jamesaimonetti:
Include https://github.com/2600hz/kazoo-crossbar/pull/514

kazoo_appex/src/kappex_controller.erl
```diff
@@ -101,8 +101,17 @@ load_app(Pid, App) ->
 init(_Args) ->
     process_flag('trap_exit', 'true'),
     _Binding = kapps_controller:binding_fetch_app(?MODULE, 'load_app', self()),
+    maybe_node_bindings(),
     {'ok', #state{}}.
 
+maybe_node_bindings() ->
+    maybe_node_bindings(kapps_config:get_is_true(?CONFIG_CAT, <<"is_aio_cluster">>, 'false')).
```
icehess: Wouldn't it better to use `<<"cluster">>` for category? I assume this is mostly useful for anything that wants to check what type of cluster is this?

kazoo_appex/src/kappex_controller.erl
```diff
@@ -101,8 +101,17 @@ load_app(Pid, App) ->
...
+
+maybe_node_bindings('false') -> 'ok';
+maybe_node_bindings('true') ->
+    lager:info("cluster running as AiO, binding for node heartbeats"),
+    kz_nodes_notify:notify_heartbeat().
```
icehess: So I guess what we want a a smooth experience for everyone, like they just need to enable the appex with less complicated steps and just starting apps.
So in that experience if someone change this settings during runtime given that the default value if `false` too, they also need to restart the kazoo_appex so bind to heartbeat.

Instead I guess it's better to always bind then move this logic of should handle the heartbeat or not to after it receives the heartbeat.

kazoo_appex/src/kappex_controller.erl
```diff
@@ -326,3 +338,33 @@ handle_app_stop({'DOWN', Ref, 'process', Pid, _Reason}
...
+    OurApps = kapps_controller:running_apps('false'),
+    _ = [maybe_start_other_app(OtherNode, AppName)
+         || {Name, _Info} <- OtherKapps,
+            AppName <- [kz_term:to_atom(Name, 'true')],
+            not lists:member(AppName, OurApps)
+        ],
+    'ok'.
+
+maybe_start_other_app(OtherNode, AppName) ->
+    case application:load(AppName) of
```
icehess: What checking for load? the appex apps for sure are not loaded if it is not started in this node.

icehess (changes requested):

icehess (approved):

PR #705, opened 2023-08-29
KZOO-418: pass config id

kusalfernando:
Previously sensitive credentials are passed under AMQP request `args` and under this change the config id is passed instead of sensitive credentials in the AMQP request. This change was originally requested by @jamesaimonetti under [pull-691](https://github.com/2600hz/kazoo-core/pull/691).


jamesaimonetti:
Can you edit the title to reflect `Ticket-Number: short description` format the project tries to maintain?

sanjeevie6353:
@kusalfernando would it be more optimum to use the term `CRM-Platform` instead of `Crm-Platform`?

kusalfernando:
> @kusalfernando would it be more optimum to use the term `CRM-Platform` instead of `Crm-Platform`?

I like that suggestion but let's wait for more opinions.

jamesaimonetti:
> > @kusalfernando would it be more optimum to use the term `CRM-Platform` instead of `Crm-Platform`?
> 
> I like that suggestion but let's wait for more opinions.

CRM is preferred as it is an acronym

kusalfernando:
@jamesaimonetti  I made the change. Could you please have a look and merge it if everything looks good?

PR #708, opened 2023-08-30
sleep before querying ledgers total after creating rollover, potentially fixes ci

icehess:
This rollover is also called for first time when we are creating the account. Db is already busy re-indexing, by querying this quickly db can lost its re-indexing and return Internal Server Error, which will cause account creation to fail. We are seeing a lot of problems in integration tests in CI failing, and in every case it is this view which cannot keep up quickly.

In account creation in CI case:
1. we create rollover for the first time
2. in ms seconds later we issue get source total query
3. at the same: a. couchdb mreview is using the pid of index process to perform the query b. couchdb also kills that index process to start a new re-indexing process killed the old indexer for modb c. But it is too late for query process because it got old indexer pid d. Query process dies with `noproc' error because old indexer is died
4. we get 500 Internal Server Error
5. account creation fails because of this

random CI tests are failing consistency in this step.


icehess:
- 5.3: https://github.com/2600hz/kazoo-core/pull/706
- 5.2: https://github.com/2600hz/kazoo-core/pull/716
- 5.1: https://github.com/2600hz/kazoo-core/pull/717

jamesaimonetti (approved):

PR #709, opened 2023-09-05
AXS-115: Start all bindings and event listeners for kappex, check config per event

manushiperera:
Previously, appex enabled and configs were checked before starting appex-client. If it was not enabled or configured, binding process was not initiated. Once configurations are updated, it it required to restart the app. 
Now, appex client is starting without checking the configs and once app fetch request is received from kapps_controller, it is checking for the configs. 
If it is enabled and configured, proceed with the request.
Otherwise ignore the request and log the error message as below. 
15:00:30.828 [error] |0000000000|kappex_controller:104(<0.162.0>) kazoo_appex is not enabled or configured properly, unable to proceed with app fetch request


kazoo_appex/src/kappex_controller.erl
```diff
@@ -95,8 +95,16 @@ stop_running(App) ->
 -spec load_app(pid(), external_app()) -> boolean().
 load_app(Pid, App) ->
     %% FIXME: maybe remove timeout from here and use timer?
-    is_process_alive(Pid)
-        andalso gen_server:call(Pid, {'load_app', App}, ?HTTP_FETCH_TIMEOUT).
+    case appex_client_config:is_enabled()
+            andalso appex_client_config:is_configured() of
+        'true' ->
+            is_process_alive(Pid)
```
jamesaimonetti: I don't think we need an aliveness test. Presumably kappex_controller is alive since the binding exists (created in init/1).

kazoo_appex/src/kappex_controller.erl
```diff
@@ -95,8 +95,16 @@ stop_running(App) ->
 -spec load_app(pid(), external_app()) -> boolean().
 load_app(Pid, App) ->
     %% FIXME: maybe remove timeout from here and use timer?
-    is_process_alive(Pid)
-        andalso gen_server:call(Pid, {'load_app', App}, ?HTTP_FETCH_TIMEOUT).
+    case appex_client_config:is_enabled()
+            andalso appex_client_config:is_configured() of
```
jamesaimonetti: Drop `of` to newline to match `case` and `end` indentation.

kazoo_appex/src/kappex_controller.erl
```diff
@@ -95,8 +95,16 @@ stop_running(App) ->
...
+        'true' ->
+            is_process_alive(Pid)
+                andalso gen_server:call(Pid, {'load_app', App}, ?HTTP_FETCH_TIMEOUT);
+        'false' ->
+            lager:error("~s is not enabled or configured properly, unable to proceed with app fetch request"
```
jamesaimonetti: Downgrade to `warning`, I don't think its a system-level error to not have the client configured properly.

jamesaimonetti (changes requested):
Overall the right direction, just a few small tweaks.

kazoo_appex/src/kappex_controller.erl
manushiperera: Done

kazoo_appex/src/kappex_controller.erl
manushiperera: Done

kazoo_appex/src/kappex_controller.erl
manushiperera: Done

kazoo_directory/src/kz_directory_device.erl
```diff
@@ -74,6 +74,7 @@ cache(EndpointId, AccountId, Profile, Origins, Options) ->
 -spec cache_callback(tuple(), kz_json:object(), callback_msg()) -> 'ok'.
 cache_callback({EndpointId, AccountId}, {CachedValue, Options}, 'erase') ->
     lager:debug("directory cache erased key (~s / ~s)", [EndpointId, AccountId]),
+    timer:sleep(kz_term:to_integer(?MILLISECONDS_IN_SECOND * 2)),
```
nkarunarathne: since MILLISECONDS_IN_SECOND = 1000 we do not need the integer conversion.

kazoo_directory/src/kz_directory_user.erl
```diff
@@ -59,6 +59,7 @@ cache(EndpointId, AccountId, Profile, Options) ->
 -spec cache_callback(tuple(), kz_json:object(), callback_msg()) -> 'ok'.
 cache_callback({EndpointId, AccountId}, {CachedValue, Options}, 'erase') ->
     lager:debug("directory cache erased key (~s / ~s)", [EndpointId, AccountId]),
+    timer:sleep(kz_term:to_integer(?MILLISECONDS_IN_SECOND * 2)),
```
nkarunarathne: As mentioned before, 
since MILLISECONDS_IN_SECOND = 1000 we do not need the integer conversion.

kazoo_appex/src/appex_client_sup.erl
```diff
@@ -47,16 +47,12 @@ start_link(HasSaneConfig) ->
 %% @end
 %%------------------------------------------------------------------------------
 -spec init(any()) -> kz_types:sup_init_ret().
-init([HasSaneConfig]) ->
+init([]) ->
```
nkarunarathne, line 50: do we need to match [] here.
Instead of that we can use _Args or something like that. right?
@jamesaimonetti  your thoughts on this?

kazoo_directory/src/kz_directory_device.erl
jamesaimonetti: This is unrelated to Manushi's changes, not sure why it showed up here

kazoo_appex/src/appex_client_sup.erl
jamesaimonetti: I prefer explicit matching when possible but in this particular case its not really a concern. Either way works for me here

PR #710, opened 2023-09-06
AXS-113 : Create payload to send to appex server

manushiperera:
Previously, OS type, OS release and OTP info were not provided to fetch a package from appex-server. But those information are important to select the most relevant package. 
Now, appex-client is providing OS type, OS release (OS name and OS version ID) and OTP info (OTP version, c compiler version, system architecture and ERTS version) as app headers to fetch the package from appex-server.

Headers [{<<"x-appex-kz-version">>,<<".0.1">>},
                      {<<"x-appex-mui-version">>,<<"5.3.1">>},
                      {<<"x-appex-os-type">>,<<"linux">>},
                      {<<"x-appex-os-release">>,<<"centos-7">>},
                      {<<"x-appex-arch-type">>,<<"x86_64-unknown-linux-gnu">>},
                      {<<"x-appex-c-compiler">>,<<"gcc-4.8.5">>},
                      {<<"x-appex-erts-version">>,<<"11.0.3">>},
                      {<<"x-appex-otp-version">>,<<"23.0.3">>},
                      {<<"accept">>,
                       <<"application/vnd.appex+kazoo, application/vnd.appex+monster, application/vnd.appex+commland">>},
                      {<<"authorization">>,<<"APPEX this is for testing">>}]


kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,77 @@ commland_version() ->
...
     file:write_file(FileName, io_lib:format("~s", [os:getpid()]), ['write', 'binary']).
+
+-spec os_type() -> kz_term:api_ne_binary().
+os_type() ->
+    try os:type() of
+        {_OSFamily, OSName} ->
+            kz_term:to_binary(OSName)
+    catch
+        E:Reason ->
+            lager:warning("retrieve of os type is resulted in an exception: ~p:~p"
```
nkarunarathne: may not need to format down this lagar since it is not exceeding 100 character together. 
Also, by putting it on one line will increase readability here.

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,77 @@ commland_version() ->
...
+
+-spec os_release() -> kz_term:api_ne_binary().
+os_release() ->
+    try os:cmd("cat /etc/os-release") of
+        OSInfoStr ->
+            OSInfoProps = [to_kv(X, "=") || X <- string:tokens(string:trim(OSInfoStr), "\n")],
+            get_os_release(OSInfoProps)
+    catch
+        E:Reason ->
+            lager:warning("retrieve of os release is resulted in an exception: ~p:~p"
```
nkarunarathne: may not need to format down this lagar since it is not exceeding 100 character together. 
Also, by putting it on one line will increase readability here.

kazoo_apps/src/kapps_util.erl
manushiperera: Done

kazoo_apps/src/kapps_util.erl
manushiperera: Done

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,74 @@ commland_version() ->
 -spec write_pid(file:filename_all()) -> 'ok' | {'error', atom()}.
 write_pid(FileName) ->
     file:write_file(FileName, io_lib:format("~s", [os:getpid()]), ['write', 'binary']).
+
+-spec os_type() -> kz_term:api_ne_binary().
+os_type() ->
+    try os:type() of
```
jamesaimonetti: How would `os:type()` crash? Under what conditions?

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,74 @@ commland_version() ->
...
+        {_OSFamily, OSName} ->
+            kz_term:to_binary(OSName)
+    catch
+        E:Reason ->
```
jamesaimonetti: Generally speaking, for variables that are only used for logging purposes, we prefix with `_` (so `_E:_Reason` in this case)

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,74 @@ commland_version() ->
...
+            lager:warning("retrieve of os type is resulted in an exception: ~p:~p", [E, Reason]),
+            'undefined'
+    end.
+
+-spec os_release() -> kz_term:api_ne_binary().
+os_release() ->
+    try os:cmd("cat /etc/os-release") of
```
jamesaimonetti: I wonder if `file:read_file/1` would be a better option (no need to invoke an OS process, run cat, and capture the result when we can just read the file contents)

jamesaimonetti: The other nice thing is you get a binary() back, which we prefer for string-y data

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,74 @@ commland_version() ->
...
+        OSInfoStr ->
+            OSInfoProps = [to_kv(X, "=") || X <- string:tokens(string:trim(OSInfoStr), "\n")],
```
jamesaimonetti: binary:split can be used once the read_file change is made

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,74 @@ commland_version() ->
...
+        E:Reason ->
+            lager:warning("retrieve of os release is resulted in an exception: ~p:~p", [E, Reason]),
+            'undefined'
+    end.
+
+-spec to_kv(string(), string()) -> {kz_term:ne_binary(), kz_term:ne_binary()}.
+to_kv(X, Separator) ->
+    [K, V] = string:tokens(string:trim(X), Separator),
+    {kz_term:to_binary(K)
+    ,kz_term:to_binary(string:trim(V, 'both', "\""))}.
```
jamesaimonetti: ending `}` on newline to match opening `}`

jamesaimonetti: See kz_binary:strip/2 for the equivalent to `string:trim`

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,74 @@ commland_version() ->
...
+-spec get_os_name(kz_term:proplist()) -> kz_term:ne_binary().
+get_os_name(OSInfoProps) ->
+    case kz_term:to_lower_binary(props:get_value(<<"ID">>, OSInfoProps)) of
+        <<"rocky">> -> <<"rockylinux">>;
+        Else -> Else
+    end.
+
+-spec get_os_version(kz_term:proplist()) -> kz_term:ne_binary().
+get_os_version(OSInfoProps) ->
+    Version = kz_term:to_float(props:get_value(<<"VERSION_ID">>, OSInfoProps)),
```
jamesaimonetti: I think we want to be more delicate here and not assume VERSION_ID is an integer or float. Look at https://www.man7.org/linux/man-pages/man5/os-release.5.html for what is considered "valid" data for this field

jamesaimonetti (changes requested):

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,74 @@ commland_version() ->
...
+-spec arch_type() -> kz_term:ne_binary().
+arch_type() ->
+    kz_term:to_binary(erlang:system_info('system_architecture')).
+
+-spec c_compiler() -> kz_term:api_ne_binary().
+c_compiler() ->
+    case erlang:system_info('c_compiler_used') of
+        {'gnuc', {A, B, C}} ->
+            kz_term:to_binary(lists:concat(["gcc-"
+                                           ,A
```
jamesaimonetti: These don't need to be on a single line each, you can combine them into one

jamesaimonetti: `<<"gcc-", (kz_binary:join([A, B, C], $.))/binary>>` could work too

kazoo_apps/src/kapps_util.erl
manushiperera: Removed the exception handling of `os:type()`

kazoo_apps/src/kapps_util.erl
manushiperera: Done

kazoo_apps/src/kapps_util.erl
manushiperera: Use `file:read_file/1` to read the file and use `kz_binary:to_proplist/3` to create the proplist

kazoo_apps/src/kapps_util.erl
manushiperera: Done

kazoo_apps/src/kapps_util.erl
manushiperera: Done

manushiperera: used `kz_binary:to_proplist/3` to create the proplist

kazoo_apps/src/kapps_util.erl
manushiperera: Done

kazoo_apps/src/kapps_util.erl
manushiperera: Removed a-z,-,_ characters of VERSION_ID and select the first numeric value as the version number.

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,56 @@ commland_version() ->
...
+        {'gnuc', {A, B, C}} ->
+            <<"gcc-", (kz_binary:join([A, B, C], $.))/binary>>;
+        Else ->
+            lager:debug("failed to retrieve c-compiler : ~p", [Else]),
+            'undefined'
+    end.
+
+-spec erts_version() -> kz_term:ne_binary().
+erts_version() ->
+    kz_term:to_binary(erlang:system_info('version')).
```
icehess: Instead of erts version get the actual OTP version. You can get the OTP version from the release folder of the running erlang:

```erlang
  -spec otp_version(bluepak_state:t()) -> undefined | binary().
  otp_version(_) ->
      case file:read_file(
               filename:join([code:root_dir(), "releases", erlang:system_info(otp_release), "OTP_VERSION"]))
      of
          {ok, Vsn} ->
              string:trim(Vsn);
          {error, Error} ->
              lager:debug("unable to detect Erlang/OTP version from its release OTP_VERSION file: ~p", [Error]),
              undefined
      end.
```

icehess: fyi this file exists since OTP 17 I guess

icehess (changes requested):

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,56 @@ commland_version() ->
...
+
+-spec get_os_version(kz_term:proplist()) -> kz_term:ne_binaries().
+get_os_version(OSInfoProps) ->
+    OSVersion = kz_binary:strip(props:get_value(<<"VERSION_ID">>, OSInfoProps), <<"\"">>),
+    OSVersionID = re:replace(OSVersion, <<"[^0-9.]">>, <<>>, ['global', {'return','binary'}]),
+    binary:split(OSVersionID, <<".">>).
+
+-spec arch_type() -> kz_term:ne_binary().
+arch_type() ->
+    kz_term:to_binary(erlang:system_info('system_architecture')).
```
icehess: fyi this doesn't really helpful at all. Erlang doesn't report proper name for all CPU types. I would use `os:cmd(uname -m)` and trim its result instead since `uname -m` at least gives us better answer.

But for now this is good enough, since x64_64 is the only supported arch for kazoo and we all assume everyone is running on that arch already.

kazoo_apps/src/kapps_util.erl
icehess: ahh, nvm, we already using `kapps_util:otp_version/0` which uses that file. that is enough then, you can drop the erts_version. OTP version already gives a a broader version which also includes erts version

kazoo_appex/src/appex_client_client.erl
```diff
@@ -225,6 +225,12 @@ fetch_app_headers() ->
...
       ,{<<"x-appex-commland-version">>, get_node_version(fun kapps_util:commland_version/0)}
       ,{<<"x-appex-meta-kazoo-version">>, get_node_version(fun kapps_util:meta_kazoo_version/0)}
+      ,{<<"x-appex-os-type">>, get_node_version(fun kapps_util:os_type/0)}
+      ,{<<"x-appex-os-release">>, get_node_version(fun kapps_util:os_release/0)}
+      ,{<<"x-appex-arch-type">>, get_node_version(fun kapps_util:arch_type/0)}
+      ,{<<"x-appex-c-compiler">>, get_node_version(fun kapps_util:c_compiler/0)}
+      ,{<<"x-appex-erts-version">>, get_node_version(fun kapps_util:erts_version/0)}
+      ,{<<"x-appex-otp-version">>, get_node_version(fun kapps_util:otp_version/0)}
       ,{<<"accept">>, kz_binary:join(
                         [<<"application/vnd.appex+", Suffix/binary>>
```
icehess: Hmm, this is wrong, can you please change it to:
```suggestion
                        [<<"application/vnd.bluepak+", Suffix/binary>>
```

icehess (changes requested):

kazoo_apps/src/kapps_util.erl
```diff
@@ -828,3 +835,56 @@ commland_version() ->
...
+            get_os_release(Props);
+        {'error', _Reason} ->
+            'undefined'
+    end.
+
+-spec get_os_release(kz_term:proplist()) -> kz_term:ne_binary().
+get_os_release(OSInfoProps) ->
+    OSID = get_os_name(OSInfoProps),
+    [OSVersionID|_] = get_os_version(OSInfoProps),
+    <<OSID/binary, "-", OSVersionID/binary>>.
```
icehess: Sorry to keep making comments repeatedly, just finding things as I go checking stuff around :) 

So the release ID format is going to be the same format we are using in our CircleCI jobs. Currently we have `build-centos7` which strip the `build-` part and we use `centos7`. We are working on Rocky Linux 8 and 9 stuff too which those get translated to `rockylinux8` and `rockylinux9` respectively.

So all said, we need to stick to the same format as this is the format appex-bluepak is using from CI environment as it publishes the packages.

So drop the `-`.
```suggestion
    <<OSID/binary, OSVersionID/binary>>.
```

icehess (changes requested):

kazoo_apps/src/kapps_util.erl
manushiperera: Done

kazoo_apps/src/kapps_util.erl
manushiperera: Done

kazoo_appex/src/appex_client_client.erl
manushiperera: [{<<"x-appex-kz-version">>,<<".0.1">>},
            {<<"x-appex-mui-version">>,<<"5.3.1">>},
            {<<"x-appex-os-type">>,<<"linux">>},
            {<<"x-appex-os-release">>,<<"centos7">>},
            {<<"x-appex-arch-type">>,<<"x86_64">>},
            {<<"x-appex-c-compiler">>,<<"gcc-4.8.5">>},
            {<<"x-appex-otp-version">>,<<"23.0.3">>},
            {<<"accept">>,
             <<"application/vnd.bluepak+kazoo, application/vnd.bluepak+monster, application/vnd.bluepak+commland">>},
            {<<"authorization">>,<<"APPEX this is for testing">>}]

kazoo_apps/src/kapps_util.erl
manushiperera: [{<<"x-appex-kz-version">>,<<".0.1">>},
            {<<"x-appex-mui-version">>,<<"5.3.1">>},
            {<<"x-appex-os-type">>,<<"linux">>},
            {<<"x-appex-os-release">>,**<<"centos7">>**},
            {<<"x-appex-arch-type">>,<<"x86_64">>},
            {<<"x-appex-c-compiler">>,<<"gcc-4.8.5">>},
            {<<"x-appex-otp-version">>,<<"23.0.3">>},
            {<<"accept">>,
             <<"application/vnd.bluepak+kazoo, application/vnd.bluepak+monster, application/vnd.bluepak+commland">>},
            {<<"authorization">>,<<"APPEX this is for testing">>}]

PR #712, opened 2023-09-11
Set twiML recodingUrl to binary type

dhuynh-Kazoo:
From the CPAAS-61 testing with TwiML `recoding`, I found that kazoo always skips storing the file because of the RecordingUrl type. `kz_http_util:urlsplit` requests the `binary` type, but the current implementation is `list` type.

Pivot PHP test file: https://dhuynh147.000webhostapp.com/twiml/record.php.


manushiperera:
@dhuynh-Kazoo, can you please use PR summary format (`previously,... now,...`) and PR title format ( `Ticket-Number: short description`) which is recommended by James?

@jamesaimonetti, this change looks fine as `kapps_call_recording:should_store_recording/2` is expecting a `kz_term:api_binary()` value for Url.

dhuynh-Kazoo:
Hi @manushiperera, thank you for your suggestions. I am checking with ops about the meta build to create a bug ticket in kazoo-pivot

PR #718, opened 2023-09-20
AXS-123 : Cleanup existing app when installing new version

manushiperera:
Previously, appex client did not clean up the old app folders, while loading newly fetched app. Therefore, it can pick up the old app folders, instead of newly fetched app. 
Now, appex client is deleting all of the old app folders related with the app name, before loading newly fetched app.


kazoo_appex/src/appex_bluepak.erl
```diff
@@ -103,6 +104,16 @@ set_load_params(#load_app{app_spec=_AppSpec
             end
     end.
 
+-spec prior_cleanup(App::atom()) -> 'ok'.
+prior_cleanup(App) ->
+    AppDir = lists:concat([App, "-*"]),
+    BasePath = code:lib_dir(),
+    DirList = filelib:wildcard(AppDir, BasePath),
+    _ = [file:del_dir_r(filename:join(BasePath, Dir))
```
icehess: I think it is good to log the directories that you are going to delete, as a warning. This may help for future debugging and also helps auditing why a certain directories was deleted.

icehess (changes requested):

kazoo_appex/src/appex_bluepak.erl
manushiperera: Warning log is added. 
`03:14:24.983 [warning] |0000000000|appex_bluepak:119(<0.23477.2>) before loading qubicle app; qubicle-0.1, qubicle-0.2, qubicle-0.3 old directory(s) is/are deleted from /home/manushi/kazoo5/_rel/kazoo/lib path`

kazoo_appex/src/appex_bluepak.erl
```diff
@@ -103,6 +104,23 @@ set_load_params(#load_app{app_spec=_AppSpec
             end
     end.
 
+-spec prior_cleanup(App::atom()) -> 'ok'.
+prior_cleanup(App) ->
+    AppDir = lists:concat([App, "-*"]),
+    BasePath = code:lib_dir(),
+    DirList = filelib:wildcard(AppDir, BasePath),
```
jamesaimonetti: Why not pattern-match the result of `filelib:wildcard/2` directly in the case instead of an additional function call to is_not empty? Something like:

```erlang
prior_cleanup(App) ->
    AppDir = lists:concat([App, "-*"]),
    BasePath = code:lib_dir(),

    case filelib:wildcard(AppDir, BasePath) of
        [] -> 'ok';
        DirList ->
            _ = [file:del_dir_r(filename:join(BasePath, Dir))
                 || Dir <- DirList
                ],
            lager:warning("before loading ~p app; ~s old directory(s) is/are deleted from ~s path"
                         ,[App, string:join(DirList, ", "), BasePath]
                         )
    end.
```

manushiperera: Done

PR #720, opened 2023-09-22
AXS-110: remove parent app tracking in kapps_controller

icehess:


icehess:
followup of https://github.com/2600hz/kazoo-core/pull/711

PR #723, opened 2023-09-26
[5.3] KZOO-410: UUID treatment for command selector

kdamas93:
Is necessary implement/automate the use of UUID for the command selector for its _id.
To do this, it is necessary to receive the uuid field in the payload through the console and in the URL from the API and this will be validated and stored in the _id and uuid fields documents.


kazoo_documents/src/kazoo_ui_apps_maintenance.erl
```diff
@@ -245,12 +245,12 @@ find_metadata(AppPath) ->
...
-            case kz_doc:id(JSON) of
-                'undefined' ->
+            case kz_json:is_defined(<<"uuid">>, JSON) of
+                'false' ->
                     {'ok', kz_doc:set_type(ValidatedJObj, kzd_applications:pvt_type())};
-                AppId ->
+                'true' ->
                     Setters = [{fun kz_doc:set_type/2, kzd_applications:pvt_type()}
-                              ,{fun kz_doc:set_id/2, AppId}
+                              ,{fun kz_doc:set_id/2, kz_json:get_value(<<"uuid">>, JSON)}
```
jamesaimonetti: You could just `get_ne_binary_value(<<"uuid">>` and use `undefined` as the false case instead.

jamesaimonetti:
Update PR title and summary to project standards

kdamas93:
> Update PR title and summary to project standards

Sorry @jamesaimonetti , I did it through cherry pick and I didn't notice, I'll correct it right now.

kazoo_documents/src/kazoo_ui_apps_maintenance.erl
kdamas93: change added !

Navoda626: @kdamas93 I think James suggesting to use `get_ne_binary_value(<<"uuid">>` in the above case in line 248. @jamesaimonetti please correct me if I'm wrong

icehess (approved):

PR #725, opened 2023-10-03
KCRO-195 : As a user I want to update date from the Monster UI entry for Marketplace connector settings

manushiperera:
Previously there was no function to return and set is_aio_cluster (all in one cluster) configuration. Now is_aio_cluster/0 and set_is_aio_cluster/1 functions are defined to return and set this configuration respectively. Further is_aio_cluster configuration value is added to the return of  get_public_configs function.


manushiperera:
Please find the related PR below. 
#https://github.com/2600hz/kazoo-crossbar/pull/522

icehess (changes requested):
Can you pelase check kazoo_appex and change places that we access `is_aio_cluster` to call this `appex_client_config:is_aio_cluster/0` function instead?

manushiperera:
Done

icehess (approved):

PR #726, opened 2023-10-05
KCRO-195: refactor get_public_configs to gracefully fetch remote configs if linked

icehess:
This a follow of previous PR  and related to crossbar pr to fetch remote cluster info from appex server if the cluster is linked.

PR #728, opened 2023-10-06
KCRO-195: set null as undefined when decoding appex server response

icehess:
This will fixes some fields that may be null on appex-server, and treat them as undefined so Kazoo will not try to aces or convert them to binary `"null"` when it is responding to UI requests.


icehess:
- related crossbar pr: https://github.com/2600hz/kazoo-crossbar/pull/522
- previous pr: https://github.com/2600hz/kazoo-core/pull/725
- previous pr: https://github.com/2600hz/kazoo-core/pull/726

kdamas93 (approved):

PR #733, opened 2023-10-13
AXS-129: actually call app.unload binding, rework load_app

icehess:
- This fixes a missing kazoo_binding:map for app.unload.
- kappex_controller is changed to only check for app fetch status, not dealing  with calling application:load anymore. If the app is fetching the process will be added to wait list to get reply back when fetch is done.
- fix kappex_controller is not replying `true` to kapps_controller after succcesful fetch


kazoo_appex/src/kappex_controller.erl
```diff
@@ -156,91 +156,21 @@ handle_call('loaded', _From, #state{loaded=Loaded}=State) ->
...
-            lager:debug("app ~s not loaded, checking listing", [ExternalApp]),
+            {'reply', 'true', State};
+        'not_fetched' ->
             handle_load_app(ExternalApp, From, State);
-        {'error', _Reason} ->
-            lager:debug("app ~s is fetched before, but it failes to load with ~p", [ExternalApp, _Reason]),
-            {'reply', 'false', State};
-        {'noreply', _} = NoReply ->
-            NoReply
+        {'fetching', #fetch{fetch=_PidRef, froms=Froms}=Fetch, NewFetchings} ->
```
jamesaimonetti, line 165: Un-matchable

kazoo_appex/src/kappex_controller.erl
```diff
@@ -323,6 +253,30 @@ code_change(_OldVsn, State, _Extra) ->
...
+    case lists:keytake(ExternalApp, #fetch.app, Fetchings) of
+        'false' ->
+            case lists:member(ExternalApp, LoadedApps) of
+                'true' ->
+                    'fetched';
+                'false' ->
+                    'not_fetched'
+            end;
+        {'value', #fetch{}, _NewFetchings}=Fetch ->
+            {'fetching', Fetch}
```
jamesaimonetti: 2-tuple here, 3-tuple expected above

manushiperera (approved):

PR #734, opened 2023-10-16
KZOO-413 : Support kazoo services specifically for 402 error modal to use in conjunction with the AppSelector API

kdamas93:
https://2600hz-commercial.atlassian.net/browse/KZOO-413

As a developer, I would like to support kazoo services specifically for 402 error modal to use in conjunction with the AppSelector API, a function will be created in kazoo services to support the 402 error modal to be used in conjunction with the AppSelector API

extra instructions about the task :

image


kazoo_services/src/kz_services_quantities.erl
```diff
@@ -505,6 +506,66 @@ app_store_updates_fold(_K, App, U) ->
...
+calculate_applications_updates(JObj, Updates) ->
+    case kz_doc:type(JObj) =:= <<"application_entitlement">> of
+        'false' -> Updates;
+        'true' ->
+            AccountId = kz_doc:account_id(JObj),
+            AppType = kz_json:get_binary_value(<<"app_type">>, JObj),
+            ViewOptions = [{'startkey', [AppType]}
+                          ,{'endkey', [AppType, kz_term:high_unicode_value()]}
+                          ],
+            case kz_datamgr:get_results(AccountId, <<"application/crossbar_listing">>, ViewOptions) of
```
icehess: When I made that comment, I was not paying attention as why this calculate function is being called and do.

The `calculate_updates` is called when you are making a change to a doc via crossbar. If you see we always recommend to use `crossbar_doc` in to deal with saving your doc. crossbar_doc is call kz_services to calculate and update the quantities passing the old doc and new (proposed) doc and kz_services can get current quantity, and check old/new doc to update the quantities. This is where `calculate_updates` is being called.

So here we have the entitlement which is already what we need. This entitlement is replacing the old apps_store doc which was a single doc to hold the apps used by the account and is users. We would use this apps_store beacuse it had all info we need to calculate how much usages this account ands it users have for each app. But here entitlement is managed by one single doc PER app, which hold this user/account usage info for us. we just need to calculate its changes (changes to the its entitlement_type). You don't need to open application (idk why i even said that tbh). the application is not something that we want to quantify here.

With all of this said, we are missing current quantity part!. If you look a bit above there is a function called `fetch_account`. This function is called when we want to get the current quantity of tings inside the account db.
This function is using `services/quantify`.

That view is special, as it doesn't exists as regular JSON file anywhere. This view is being generated dynamically by `kazoo_services_maintenance:register_views/0`. You need to update `quantify` view to also handle `application_entitlement` doc type.

icehess (changes requested):

kazoo_services/src/kz_services_quantities.erl
icehess: I slightly updated the above comment to fix some terminologies.

kdamas93: @icehess changes added to the view (`services/quantify`) , I would appreciate your comments on this change.

kazoo_services/src/kazoo_services_maintenance.erl
```diff
@@ -214,6 +214,20 @@ js_quantify_map() ->
        "                emit(['account_apps', doc.apps[index].name], 1);"
        "            }"
        "            break;"
+       "        case 'application_entitlement':"
+       "            switch (doc.type) {"
+       "                case 'all':"
+       "                    emit(['user_applications', doc.app_id], -1);"
```
icehess: Hmm I'not sure we want to introduce a new service item for this, we can use the same `user_apps` and `account_apps`. Ask Karl on what this should be.

kazoo_services/src/kz_services_quantities.erl
icehess: okay you only addressed the view, what about this function here? you don't need to open any apps, just check the entitlement type, that is all.

icehess (changes requested):

kazoo_services/src/kz_services_quantities.erl
kdamas93: @icehess change added ! ty!

icehess (approved):

PR #736, opened 2023-10-20
KZOO-413 : Support kazoo services specifically for 402 error modal to use in conjunction with the AppSelector API

kdamas93:
This performs the treatment for the individual charge of applications which was reported in this video:

image

https://github.com/2600hz/kazoo-core/assets/68119098/cbdb9b07-8e2c-4de4-b1fe-6a29f8558edf


kazoo_services/src/kz_services_plans.erl
```diff
@@ -549,6 +550,27 @@ editable_monster_apps_fields(Fields, ResellerId) ->
             ]
     end.
 
+-spec editable_commland_apps_fields(kz_term:proplist(), kz_term:api_binary()) -> kz_term:proplist().
+editable_commland_apps_fields(Fields, 'undefined') ->
+    {'ok', MasterAccountId} = kapps_util:get_master_account_id(),
+    editable_commland_apps_fields(Fields, MasterAccountId);
+editable_commland_apps_fields(Fields, ResellerId) ->
+    AppType = <<"commland_ui">>,
```
icehess, line 558: Per our discussion,  this must all applications type! the applications is the future for both MUI and commland and any that comes in the future!

kazoo_services/src/kz_services_plans.erl
```diff
@@ -549,6 +550,27 @@ editable_monster_apps_fields(Fields, ResellerId) ->
...
+    ViewOptions = [{'startkey', [AppType]}
+                  ,{'endkey', [AppType, kz_term:high_unicode_value()]}
+                  ],
+    case kz_datamgr:get_results(ResellerId, <<"application/crossbar_listing">>, ViewOptions) of
+        {'ok', JObjs} ->
+            Applications = [kz_doc:id(Application)|| Application <- JObjs],
```
icehess, line 564: ~use `kzd_serviceplan_item:name/2` with `kz_doc:id/1` as default value im case there is no name~

kazoo_services/src/kz_services_plans.erl
```diff
@@ -549,6 +550,27 @@ editable_monster_apps_fields(Fields, ResellerId) ->
...
+            [{<<"account_commland_apps">>, Applications}
```
icehess, line 565: again don't use introduce a new service category! use the old one `account_apps` and `user_apps`!

icehess (changes requested):

kazoo_services/src/kz_services_plans.erl
icehess: aah, nvm. this the application doc , not service plan.

there are multiple options:

1. return id as you do now, then UI needs to make subsequent API call to fetch the app metadata from API which may be come very cumbersome!
2. this application doc so you have name of the app, then return its name!
3. You can work with UI team to let them know they may expect a list of strings (the old school way) or a mix list of string and/or objects  (the new way) which consist of both the app id and app name

jamesaimonetti:
reopen with suggested changes if still relevant

PR #742, opened 2023-11-07
KZOO-423: delete yodb when deleting account too

icehess:
This adds deleting account's YODBs when deleting the an account.
Also adding `kzdb_accounts:force_cleanup_account/1` to help deleting lingering [yo/mo]dbs even if account's doc is missing.


jamesaimonetti:
Please rebase and reopen if still relevant

PR #747, opened 2023-11-16
[5.3] KZOO-418: pass config id

kusalfernando:
Previously sensitive credentials are passed under AMQP request `args` and under this change the config id is passed instead of sensitive credentials in the AMQP request.


icehess (approved):

kusalfernando:
Master branch PR: https://github.com/2600hz/kazoo-core/pull/705

PR #753, opened 2023-12-06
[master] KCRO-200: Extension Reassignment Error Leads to Unintended Overwrite

filipevarjao:
We moved the validation for presence id from crossbar to `validate/3` function at core so we can validate it for PUT/PATCH/POST. 
[test](https://github.com/2600hz/kazoo-properly/pull/191)


kazoo_documents/src/kzd_users.erl
```diff
@@ -94,6 +94,7 @@
 
 -define(LIST_BY_USERNAME, <<"users/list_by_username">>).
 -define(LIST_BY_HOTDESK_ID, <<"users/list_by_hotdesk_id">>).
+-define(LIST_MAILBOX, <<"vmboxes/listing_by_mailbox">>).
```
lazedo: what's the purpose of this ? user may not have a vmbox but still have a presence_id assigned

lazedo (commented):
clarify on the vmboxes/listing_by_mailbox usage

kazoo_documents/src/kzd_users.erl
filipevarjao: ok

kazoo_documents/src/kzd_users.erl
```diff
@@ -1051,6 +1053,37 @@ do_validation(AccountId, UserId, ReqJObj, ValidateFuns) ->
                ,ValidateFuns
                ).
 
+%%------------------------------------------------------------------------------
+%% @doc If set, Validate the user's username is unique within the account.
+%% @end
+%%------------------------------------------------------------------------------
+-spec maybe_validate_presence_uniqueness(kz_term:api_ne_binary(), kz_term:api_ne_binary(), kazoo_documents:doc_validation_acc()) ->
+          kazoo_documents:doc_validation_acc().
+maybe_validate_presence_uniqueness(AccountId, UserId, {Doc, Errors}) ->
```
jamesaimonetti, line 1062: I'd love to see a properly test that shows what the issue is. I am not aware of a requirement that presence_id is unique to a user in an account, so at a minimum I would expect this check to be opt-in on the account/cluster level?

Thoughts on that?

jamesaimonetti (changes requested):

kazoo_documents/src/kzd_users.erl
```diff
@@ -1051,6 +1053,37 @@ do_validation(AccountId, UserId, ReqJObj, ValidateFuns) ->
...
+        NewPresence ->
+            MailboxList = kz_datamgr:get_results(AccountId, ?LIST_USER_CB),
+            check_uniqueness(NewPresence, MailboxList, {Doc, Errors})
+    end.
+
+-spec check_uniqueness(kz_term:ne_binary(), kazoo_data:get_results_return(), kazoo_documents:doc_validation_acc()) -> kazoo_documents:doc_validation_acc().
+check_uniqueness(_, {'error', _}, {Doc, Errors}) -> {Doc, Errors};
+check_uniqueness(_, {'ok', []}, {Doc, Errors}) -> {Doc, Errors};
+check_uniqueness(NewPresence, {'ok', [CbUser|CbUsers]}, {Doc, Errors}) ->
+    case kz_json:get_value([<<"value">>, <<"presence_id">>], CbUser) of
```
jamesaimonetti: use typed accessors

kazoo_documents/src/kzd_users.erl
filipevarjao: https://github.com/2600hz/kazoo-properly/pull/191

jamesaimonetti:
Holding merge until we figure out whether this change should be added to KAZOO as it introduces a backwards incompatible requirement that "presence_id" be unique for users.

filipevarjao:
Decided to not presence_id to be unique.

PR #754, opened 2023-12-11
KTAB-152: add more flexible `gen_listener` binding types

danielfinke:
- used for specifying tighter constraints on KAPI binding props per module
- update types in `gen_listener` to use `binding_module()`, `binding_properties`, etc.
- fix crash in `gen_listener:rm_queue` resulting from bad match on `other_queues`


kazoo_amqp/src/gen_listener.erl
```diff
@@ -1316,14 +1320,14 @@ remove_binding(Binding, Props, Q) ->
...
--spec create_binding(kz_term:ne_binary(), kz_term:proplist(), kz_term:api_ne_binary()) -> any().
+-spec create_binding(binding_module(), binding_properties(), kz_term:api_ne_binary()) -> any().
 create_binding(Binding, Props, QueueName) when not is_binary(Binding) ->
     create_binding(kz_term:to_binary(Binding), Props, QueueName);
 create_binding(Binding, Props, QueueName) ->
     %% TODO
     %% HANDLE POTENTIAL CRASH
     %%
-    APIModule = kz_term:to_atom(list_to_binary([<<"kapi_">>, kz_term:to_binary(Binding)]), 'true'),
+    APIModule = kz_term:to_atom(list_to_binary([<<"kapi_">>, Binding]), 'true'),
```
jamesaimonetti: What about `kz_term:to_atom(<<"kapi_", Binding/binary>>, 'true')` ? I assume Binding is a binary() in this clause, right?

danielfinke: @jamesaimonetti yep, that's nicer. Just didn't look at it too much when I made the change 😄

PR #758, opened 2023-12-27
KTEL-21: As a reseller I want my emails to be signed with DKIM so that they are reliably delivered

manushiperera:
Previously, DKIM information is not used while sending emails. Now, file id of private key file of DKIM is saved in `whitelabel` document as a `pvt` field. Further private key file is saved separately as an extra precaution.` fetch_private_key_id/2` function is used to retrieve the file id of private key file.


manushiperera:
Related PRs:

https://github.com/2600hz/kazoo-crossbar/pull/545
https://github.com/2600hz/kazoo-teletype/pull/70

kazoo_documents/src/kzd_whitelabel.erl
```diff
@@ -409,3 +423,16 @@ fetch_port_authority(Account, Default) ->
             lager:debug("failed to open whitelabel doc for ~s", [Account]),
             Default
     end.
+
+-spec fetch_private_key_id(kz_term:api_binary(), Default) -> kz_term:api_ne_binary() | Default.
```
jamesaimonetti: We probably want to put the setter version in this module too.

I would also update the name of this function. In general, the `getter` is the name of the key, while the `setter` is `set_{KEY}`.

In this case, I would make nice names like `dkim_doc_id` or `dkim_key_id` and `set_dkim_doc_id` or `set_dkim_key_id`.

manushiperera: Done

PR #767, opened 2024-01-04
Revert "KPUS-7: add Authorizing ID and Type into missed_call payload- schema correction"

lazedo:
Reverts 2600hz/kazoo-core#765

that is the question, why was this requested ? what did verizon actually requested ? it doesn't make sense to include this information in the payload, afaics

  [6 days ago](https://2600hz.slack.com/archives/CH9JJLZED/p1703867467656569?thread_ts=1703865516.159049&cid=CH9JJLZED)


jamesaimonetti:
@lazedo sorry, missed that thread in Slack.

@wdhammika can you weigh in on the context for why this info should be added?

lazedo:
my bad @jamesaimonetti , i should have commented on the PR not on slack

k-anderson:
@lazedo the VZ specific use case was because their client would be using multiple devices.  As such when a push notification came in they wanted to be able to differentiate which device had caused the push to be able to suplement the UX prior to establishing the call.  What is the concern adding this information?

PR #769, opened 2024-01-05
KZOO-432: CRM platform can be undefined and account ID is required

kusalfernando:
In addition to supporting passing CRM platform ID for the AMQP calls, we also need to support the current queries with no ID. Also, the account ID must pass with the AMQP request to maintain account-wise configs.


danielfinke (approved):

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -157,14 +157,13 @@ caller_info_definition() ->
...
-                                                            ,<<"CRM-Platform">>
                                                             ]}
-              ,{fun kapi_definition:set_optional_headers/2, []}
+              ,{fun kapi_definition:set_optional_headers/2, [<<"CRM-Platform">>
+                                                            ]}
               ,{fun kapi_definition:set_values/2
                ,kapi_definition:event_type_headers(Category, EventName)
                }
               ,{fun kapi_definition:set_types/2, [{<<"Caller-ID">>, fun kz_term:is_ne_binary/1}
-                                                 ,{<<"CRM-Platform">>, fun kz_term:is_ne_binary/1}
```
danielfinke, line 167: I *think* you can still have the type check in here with the header being optional

kazoo_amqp/src/api/kapi_crm.erl
```diff
@@ -221,7 +221,7 @@ interaction_definition() ->
                                                  ,{<<"Contact-ID">>, fun kz_term:is_ne_binary/1}
                                                  ,{<<"Direction">>, fun kz_term:is_ne_binary/1}
                                                  ,{<<"Disposition">>, fun kz_term:is_ne_binary/1}
-                                                 ,{<<"CRM-Platform">>, fun kz_term:is_ne_binary/1}
+                                                 ,{<<"CRM-Platform">>, fun kz_term:is_api_binary/1}
```
danielfinke: hey, out of curiosity, since we made the header optional, does it work to leave this as it was before I requested the change to `is_api_binary`?
```suggestion
                                                 ,{<<"CRM-Platform">>, fun kz_term:is_ne_binary/1}
```

kusalfernando: yep, that was a mistake. I will fix that.

PR #770, opened 2024-01-10
KZOO-448 Client is unable to add/remove apps on a specific account

nkarunarathne:
Previously, there weren't a default value for 'allowed_users'

Now, added 'specific' as default value. This will handled the crash when 'allowed_users' field is 'undefined'.

https://oomacorp.atlassian.net/browse/KZOO-448


k-anderson:
It is not expected that allowed_users ever be empty, hence no default but I don't have anything against defensively handling this in the event the data is unexpected.  However, I don't think the default should be admins.  Looking at the enforcement code here https://github.com/2600hz/kazoo-crossbar/blob/master/src/cb_apps_util.erl#L583 that defaults to "specified" and if there is no list of users it just assumes an empty list.  The quantifier should mirror that behavior so the logic for what we charge matches the logic of how we enforce access.

k-anderson:
@nkarunarathne looks good to me, need need CI to pass :)

PR #771, opened 2024-01-12
KBLA-13: VMbox ID not emitted on object.doc_created.mailbox_message

Navoda626:
Previously for the blackhole event `object.doc_created.mailbox_message`, `source_id` and `source_type` parameters which are available in the `mailbox_message` document were not included.

Now if source_id` and `source_type` parameters available in doc update event they will include in the blackhole event.


kazoo_data/src/kzs_publish.erl
```diff
@@ -170,6 +170,8 @@ publish(Action, Db, Doc) ->
               ,{<<"Date-Modified">>, kz_doc:created(Doc)}
               ,{<<"Date-Created">>, kz_doc:modified(Doc)}
               ,{<<"Is-Soft-Deleted">>, IsSoftDeleted}
+              ,{<<"Source-Id">>, kz_json:get_binary_value(<<"source_id">>, Doc, 'undefined')}
```
jamesaimonetti: Use `get_ne_binary_value` unless an empty string is a useful value to have.

Generally "empty" values like [], <<>>, or 'undefined', are removed from payloads. Also, `'undefined'` is the default when you use arity-2 so no need to specify it here.

icehess (commented):
What do you think about adding `/vmboxes/message/{MESSAGE_ID}` to `cb_vmboxes` instead of this change?

(also same for cb_faxboxes maybe)
also cc @jamesaimonetti

jamesaimonetti:
@icehess can you explain how your request relates to this PR? I don't follow

icehess:
@jamesaimonetti the Jira ticket is about the blackhole event for vm message doc created doesn't have the the actual vmbox id, so they cannot fetch it from API. cb_vmboxes currently requires both vmbox id and message id.

I'm suggesting that we can probably add a new path to cb_vmbxoes without requiring vmbox id.

icehess:
welp i depends on the user privileges i guess, meaning fetching by only message id needs to be restricted to only admins.

Anyhow, imho adding more payload to to doc update event is not good idea.

kazoo_data/src/kzs_publish.erl
Navoda626: Got it. This is updated

jamesaimonetti:
@icehess i see, thanks for the clarification.

I do think having the vmbox ID on the blackhole event is worthwhile. I could see doing like we did with webhook payloads where "v2" gets a fuller payload delivered.

Could we indicate on the websocket subscription for object events to include full doc?

Navoda626:
Closing this PR since this solution is not valid

PR #777, opened 2024-01-31
KZOO-341: move `kazoo_hooks` to use pooled bindings

danielfinke:
Add support for channel event filters when binding for pooled call events in `kapi_call`. Each filter is registered in `gproc` individually. Can mix pooled/non-pooled bindings in `restrict_to`.
  - ensure wildcard for call ID or event name works when events are published

Deprecate `kz_hooks:bind` in favour of `kz_events` async event handlers. Async handlers provide no guarantee of event ordering, like the legacy `kz_hooks:bind` behaviour.
  - fix bug where bindings would not be triggered unless `kz_hooks_listener` bindings were started separately

Mark `kazoo_hooks` modules as deprecated in favour of `kz_events` in general.

Simplify `kz_event_listener` presence bindings by exporting and using the KAPI definition.

## Related PRs

- 2600hz/kazoo-bluejeans#16
- 2600hz/kazoo-webhooks#46


danielfinke:
Review discussion thread: https://2600hz.slack.com/archives/CH9JJLZED/p1706810391017939

lazedo (approved):

PR #778, opened 2024-02-06
KCRO-204: As an admin I want the ability to create a JWT that only grants read access

manushiperera:
Previously, it was unable to grant read only access for specific users. Now, if `pvt_scopes` field is defined in `user` doc with `fetch-exclusive`, creates the JWT token that contains a claim for read only access. When JWT is authorized, check for that claim and if it presents then the request is denied if the HTTP method is not a `GET`.


manushiperera:
Related PRs
https://github.com/2600hz/kazoo-crossbar/pull/549
https://github.com/2600hz/kazoo-properly/pull/196

kazoo_auth/src/kz_auth_claims.erl
```diff
@@ -46,6 +47,13 @@ meta_id(Claims) ->
 identity_sig(Claims) ->
     kz_json:get_ne_binary_value(<<"identity_sig">>, Claims).
 
+-spec scope(claims()) -> kz_term:api_ne_binary().
+scope(Claims) ->
+    scope(Claims, 'undefined').
+
+-spec scope(claims(), Default) -> kz_term:ne_binary() | Default.
+scope(Claims, Default) ->
+    kz_json:get_ne_binary_value(<<"scope">>, Claims, Default).
```
lazedo, line 56: there's no `scope` , do you mean `scopes` ?

manushiperera: In here, it says `scope`
https://datatracker.ietf.org/doc/html/rfc8693#name-scope-scopes-claim

Please correct me, if I misunderstood.

lazedo: you're right, my bad, `scopes` is the internal usage that is done on kz_auth_scope (has , available, ...). still don't understand why is this needed ?

manushiperera: Ooma wants to access the financial data of `zswitch` urgently. So `fetch_exclusive` scope is introduced to make sure that the users could not manipulate the data. Since `scope_restrictions` are yet to be developed, `pvt_scopes` in `user` doc is used to cater this urgent internal requirement.

lazedo: but have you seen this https://github.com/2600hz/kazoo-crossbar/blob/master/doc/scope_restrictions.md and this https://github.com/2600hz/kazoo-crossbar/blob/master/doc/scopes.md ? if not, I would recommend to go thru this. if you have, where does this fail to work ?

manushiperera: It shows `not found` error for `scope_restrictions` APIs. Seems that the request is not received by `cb_scope_restrictions` module.

lazedo: did you ensure the module was started ? 
also look to token_restrictions module and doc. I feel this and related PRs are not needed

manushiperera: After starting the module, I was able to update `user` doc with `scope_restrictions` and the `scope`  is available in the JWT payload.
```
payload =>
               #{<<"account_id">> => <<"97eed49fdd774dda2d3490af6e1aa99b">>,
                 <<"exp">> => 1707763977,
                 <<"identity_sig">> =>
                     <<"sklggElqf2JULNu-fnX3mM9ZThzUKkZnEn5Ddp7AhmE">>,
                 <<"iss">> => <<"kazoo">>,<<"method">> => <<"cb_user_auth">>,
                 <<"owner_id">> => <<"1f951ab0133a23611a0c756add48a222">>,
                 <<"scope">> => <<"crossbar:read_only">>},
```

In api_util:is_permitted_scope/2, it is only checking whether the requested scopes are present in the token and not executing scope restrictions over the API request. Therefore all the API requests are passing. 

@k-anderson can you please assist on this?

icehess: The intended endpoints must implement `allowed_scopes/0,1` callback.

https://github.com/2600hz/kazoo-crossbar/blob/master/doc/scopes.md#crossbar-binding

lazedo: you need to enable the token restrictions and configure them by account and/or system

lazedo: @icehess `intended endpoints implementation` only needed if the module needs to do special handling not handled by rules

icehess: @lazedo I don't see how scopes in Claim are enforced. The only place I see is at `api_util:enforce_scopes/3` and that is working on whatever `allowed_scopes/1` binding returns.

https://github.com/2600hz/kazoo-crossbar/blob/master/src/api_util.erl#L2177

icehess: All those configurations are just about what to set as scopes in token's Claim. Nothing in the code is enforcing scoped set in token other than that `enforce_scopes`

lazedo: @icehess cb_token_restrictions binds to authorize and handles that

manushiperera: Seems that it it retrieving and merging both `scope_restrictions` and `token_restrictions` while checking for  auth_restrictions.
https://github.com/2600hz/kazoo-crossbar/blob/master/src/modules/cb_token_restrictions.erl#L190
Even though I configured both `scopes` and `scope_restrictions` , it received empty doc for `scope_restrictions` and unable to restrict via scopes.
https://github.com/2600hz/kazoo-crossbar/blob/master/src/crossbar_util.erl#L1032

But configuring only below `token_restriction` rule in the account via `token_restrictions` API , it grants read only access and other APIs are rejected with `403 forbidden error`.

```
 {
    "data": {
        "restrictions": {
            "cb_user_auth": {
                "user": {
                    "_": [
                        {
                            "rules": {
                                "*": [
                                    "GET"
                                ]
                            }
                        }
                    ]
                }
            }
        }
    }
}
```

Thanks @lazedo

manushiperera:
Since we can use existing `scope_restriction` to cater this requirement, this PR is irrelevant.

First, it is required to define a scope restriction along with token restriction rule.
```
  "scopes": [
    "crossbar:read_only"
  ],
  "token_restrictions": {
    "_": [
      {
        "allowed_accounts": [
          "_"
        ],
        "rules": {
          "#": [
            "GET"
          ]
        }
      }
    ]
  }
```
Then we can use that doc id to define `scope_restrictions` in user doc.

```
{
        "last_name":"test1",
        "first_name": "test0",
        "password": "test1",
        "username": "test1",
        "scope_restrictions": [SCOPE_RESTRICTION_DOC_ID]
}
```

PR #784, opened 2024-02-21
[5.3] KTEL-21: As a reseller I want my emails to be signed with DKIM so that they are reliably delivered

manushiperera:
Previously, DKIM information is not used while sending emails. Now, file id of private key file of DKIM is saved in `whitelabel` document as a `pvt` field. Further private key file is saved separately as an extra precaution. `fetch_private_key_id/2` function is used to retrieve the file id of private key file.


manushiperera:
Related PRs

https://github.com/2600hz/kazoo-teletype/pull/71
https://github.com/2600hz/kazoo-crossbar/pull/551
https://github.com/2600hz/kazoo-properly/pull/197

icehess (approved):

PR #785, opened 2024-02-27
KZOO-449: Skip number services when previous account doc not exist

dhuynh-Kazoo:
When getting services from a previously assigned account of a number, there may be a case that the account doc does not exist in the account db anymore which will result in a crash.

Fixing the crash by skipping getting services from this account.


kazoo_numbers/src/knm_ops.erl
```diff
@@ -283,15 +283,21 @@ run_services([AccountId|AccountIds], Updates, UpdatedServicesAcc) ->
...
-    UpdatedServices = kz_services:set_updates(Services
-                                             ,AccountId
-                                             ,CurrentJObjs
-                                             ,ProposedJObjs
-                                             ),
-    Quotes = kz_services_invoices:create(UpdatedServices),
-    HasAdditions = kz_services_invoices:has_billable_additions(Quotes),
-    check_creditably(Services, Quotes, HasAdditions),
-    run_services(AccountIds, Updates, [UpdatedServices | UpdatedServicesAcc]).
+    case kz_services:services_jobj(Services) of
```
icehess: Could you add a comment above this line that describe you are doing this because `services_group_number/3` may return the account id of the account this number was previously assigned, and since then that account is maybe removed.

Also update the PR description to reflect this fact too.

icehess (changes requested):

kazoo_numbers/src/knm_ops.erl
dhuynh-Kazoo: hi @icehess, I updated the comment

kazoo_numbers/src/knm_ops.erl
```diff
@@ -283,15 +283,22 @@ run_services([AccountId|AccountIds], Updates, UpdatedServicesAcc) ->
...
-    UpdatedServices = kz_services:set_updates(Services
-                                             ,AccountId
-                                             ,CurrentJObjs
-                                             ,ProposedJObjs
-                                             ),
-    Quotes = kz_services_invoices:create(UpdatedServices),
-    HasAdditions = kz_services_invoices:has_billable_additions(Quotes),
-    check_creditably(Services, Quotes, HasAdditions),
-    run_services(AccountIds, Updates, [UpdatedServices | UpdatedServicesAcc]).
+    %% skip updating services if account id doc could not be found
```
icehess:
```suggestion
    %% Skip updating services if account id doc could not be found.
    %% This should be only be the case if previous assigned account is deleted.
```

icehess (changes requested):
Hi @dhuynh-Kazoo, can you check this, please reach out to Ops and ask if they can pull down the number doc from db directly and check if `pvt_previously_assigned_to` is set and it is this ID that causing the crash in services.

icehess (changes requested):
~Thinking of this, @dhuynh-Kazoo instead of relying on service jobj is undefined. lets check if it is undefined do another hard check if `kz_datamgr:open_doc(AccountId, AccountId)` id it returns `{'error', 'not_found'}` then consider that the account is deleted, log this and continue. 
for any else crash as usual by run service failed for this account.~

Ignore this

dhuynh-Kazoo:
THank you, I will check it

icehess:
@dhuynh-Kazoo With the recent look at the ticket and number doc, it seems the account id in `pvt_assigned_to` is missing.
So revert this change.

Checking previously assigned is still useful, so instead of check for undefined service doc, change last function clause of `services_group_number/4` and do `kz_datamgr:open_doc(AccountId, PrevAssignedTo)` and ignore it ONLY if `{error, not_found}`.
(We only care about PrevAssignedTo in last clause, other cases and also AssginedTo must strictly exists as before and should result of crash as before if the service runs fails for them).

kazoo_numbers/src/knm_ops.erl
```diff
@@ -390,11 +390,18 @@ services_group_number(PhoneNumber, AssignedTo, AssignedTo) ->
 services_group_number(PhoneNumber, AssignedTo, PrevAssignedTo) ->
     ProposedJObj = knm_phone_number:to_json(PhoneNumber),
     CurrentJObj = knm_phone_number:current_doc(PhoneNumber),
-    [{[AssignedTo, <<"proposed">>], ProposedJObj}
-    ,{[AssignedTo, <<"current">>], kz_json:new()}
-    ,{[PrevAssignedTo, <<"proposed">>], kz_json:new()}
-    ,{[PrevAssignedTo, <<"current">>], CurrentJObj}
-    ].
+    case kzd_accounts:fetch(PrevAssignedTo) of
```
dhuynh-Kazoo: hi @icehess, I updated the changes. It may differ from your idea (or maybe because I do not get you well).

- For the first option, I understand from your suggestion _kz_datamgr:open_doc(AccountId, PrevAssignedTo)_, we should use the _AccountId_ from _CrossbarOptions_. And we do not have `AccountId` passed in _run_services/1_ (yeah we can get Options from the map). But for the case we do not have `AccountId` in _CrossbarOptions_, how do we handle that for run_services/1?

- For the second option, I understand that you refer to _kzd_account:open_cache_doc(AccountId, AccountId)_ to check the account doc in account db then we skip _PrevAssignedTo_ if it is not_found.
Regarding this method, I am concerned about the case where we can get ServicesJObj from services db. Still, the account doc is not in the account db so somehow we skip the _PrevAssignedTo_ ( it is incorrect from the current implementation, check the existing of account doc in service first) -> **my latest change**. 
I am thinking that keep checking _undefined ServiceJObs_ to avoid missing first check (account doc in services db).

kazoo_numbers/src/knm_phone_number.erl
```diff
@@ -822,7 +822,7 @@ prev_assigned_to(#knm_phone_number{prev_assigned_to=PrevAssignedTo}) ->
 -spec set_prev_assigned_to(pn_record(), kz_term:api_ne_binary()) -> pn_record().
 set_prev_assigned_to(PN=#knm_phone_number{prev_assigned_to = 'undefined'}
                     ,PrevAssignedTo=?MATCH_ACCOUNT_RAW(_)) ->
-    PN#knm_phone_number{prev_assigned_to = PrevAssignedTo};
+    PN#knm_phone_number{prev_assigned_to = maybe_set_prev_assigned_to(PrevAssignedTo)};
```
icehess: Confused why change here, and why not do the same check for other clauses below?

I don't think this is a good change here. Yes the account may gone but still a piece of history of the number we want to store in number doc.

I still think this is best to be handle in [`services_group_numbers/2` function](https://github.com/2600hz/kazoo-core/blob/master/kazoo_numbers/src/knm_ops.erl#L358-L360).

For example:

```erlang
    PrevAssignedTo = validate_prev_assigned_to(
                       knm_phone_number:prev_assigned_to(PhoneNumber)
                      ),
```

Then:

```erlang
validate_prev_assigned_to('undefined') ->
    'undefined';
validate_prev_assigned_to(PrevAssginedTo) ->
    case kz_datamgr:open_doc(PrevAssginedTo, PrevAssginedTo) of
        {'error', 'not_found'} -> 'undefined';
        _ -> PrevAssginedTo
    end.
```

I used `kz_datamgr:open_doc/2` here because cache may be stale. I don't have a strong opinion against using `open_cache_doc`/

icehess (changes requested):

kazoo_numbers/src/knm_ops.erl
icehess: Sorry still not sure what are you saying here. I didn't ask for `kz_datamgr:open_doc(AccountId, PrevAssignedTo)` I asked only for `kz_datamgr:open_doc(PrevAssignedTo, PrevAssignedTo)`.

kazoo_numbers/src/knm_phone_number.erl
```diff
@@ -1864,3 +1864,28 @@ add_on_success(PN = #knm_phone_number{on_success = CB}, Callback) ->
 -spec reset_on_success(pn_record()) -> pn_record().
 reset_on_success(PN = #knm_phone_number{}) ->
     PN#knm_phone_number{on_success = []}.
+
+-spec maybe_set_prev_assigned_to(kz_term:ne_binaries()) -> 'undefined' | kz_term:ne_binaries().
+maybe_set_prev_assigned_to(PrevAccountId) ->
+    case
+        is_account_service_doc_exist(PrevAccountId)
```
icehess: See above note, but even this change is not necessary, checking only the account exists is enough! don't mess with service doc explicitly anywhere other than services. kazoo services knows how to handle missing service doc.

icehess (changes requested):

kazoo_numbers/src/knm_ops.erl
dhuynh-Kazoo: hi @icehess, I thought that 
```
kzd_accounts:fetch(PrevAssignedTo) -> open_cache_doc(AccountId, AccountId) -> kz_datamgr:open_cache_doc(Db, AccountId, Options). 
```
It's similar to your suggestion and I referred to `kz_services:create/1` so I selected that.

But I will apply your suggestion
```
kz_datamgr:open_doc(PrevAssignedTo, PrevAssignedTo)
```

kazoo_numbers/src/knm_phone_number.erl
dhuynh-Kazoo: I got your idea. Thank you!

icehess (approved):

icehess (approved):

PR #787, opened 2024-02-29
KWEB-12: replace `kapi_conf:doc_type_update` with `kapi_webhooks`

danielfinke:
- `doc_type_update` specific to auto-disabled webhook re-enablement replaced with a dedicated KAPI (`kapi_webhooks`)
- still possible to bind for doc change events for a specific doc type, using, e.g. for `user` docs:

  ```erlang
  {'conf', [{'restrict_to', ['doc_updates']}
           ,{'type', <<"user">>}
           ]}
  ```

  - in fact, `doc_type_update` never served this purpose, as it was only used for re-enabling webhooks, and not for routing doc change events
- update `conf` pooled binding routing key to `#`

## Related PRs

- 2600hz/kazoo-crossbar#553
- 2600hz/kazoo-webhooks#47


lazedo:
@danielfinke since you're touching this, can you add the handling of `types`  as alternative to `type` ?

```
{'conf', [{'restrict_to', ['doc_updates']}
         ,{'types', [<<"user">>, <<"device">>]}
         ]}

```

danielfinke:
@lazedo sure thing, but do you mind if I separate that into another PR focused on adding that prop?

danielfinke:
@lazedo done in #788

PR #798, opened 2024-03-12
KZOO-447: add a maintenance command to report mismatch pvt_reseller_id

icehess:
This adds a few new maintenance commands to report and fix wrong
`pvt_reseller_id` in service docs.

In an unrelated issue, the service doc can end up in state of having wrong
`pvt_reseller_id`. We didn't find why/how but we suspect it was due to
moving account to be under another account and it caused setting the
`pvt_reseller_id` to Super Duper account instead.

This caused the sub-account to be billed directly if master account has
different bookkeeper than default or reseller.

This PR adds maintenance commands to first get report and/or fix service
docs by comparing the first reseller id in `pvt_tree` and the value of
`pvt_reseller_id`.

For doing so, a list of reseller will be fetched from `accounts` db and crawling
service docs. If a service doc `pvt_tree` has an intersection of list of reseller and the
last item is not equal to `pvt_reseller_id` the account gets reported.

After confirmation, a sysadmin can run another maintenance command to repeat
the same process but actually fix and correct the `pvt_reseller_id`.


jamesaimonetti (approved):

icehess:
- Sup command PRs (core):
    - master: https://github.com/2600hz/kazoo-core/pull/798
    - 5.3: https://github.com/2600hz/kazoo-core/pull/802
    - core fix-5.3.20: https://github.com/2600hz/kazoo-core/pull/803
- Fix move account command PRs (crossbar):
    - master: https://github.com/2600hz/kazoo-crossbar/pull/557
    - 5.3: https://github.com/2600hz/kazoo-crossbar/pull/558
    - crossbar 5.3.11: https://github.com/2600hz/kazoo-crossbar/pull/559

PR #801, opened 2024-03-15
[master] KPIV-9: Set recordingSid with media name

dhuynh-Kazoo:
Currently for twiml format in the pivot application, the `RecordingSid` is not included in the query list when sending the recorded call to the Url in `action` attribute.That cause missing this field. 
So just set MediaName to `recording_sid` of call object.


sanjeevie6353:
Reviewed...

kazoo_translator/src/converters/kzt_twiml.erl
```diff
@@ -404,6 +404,7 @@ finish_record_call(Call, Props, MediaName) ->
 
                 _ = kapps_call_command:store_file(filename:join([<<"/tmp">>, MediaName]), StoreUrl, Call),
                 [{fun kzt_util:set_recording_url/2, StoreUrl}
+                ,{fun kzt_util:set_recording_sid/2, MediaName}
```
icehess, line 407: `RecordingSid` is actually must be the Document ID of the record metadata we save in the database.

Lets pause here, there are multiple issue with this code, first of all it is old and we may want to do things a bit differently these days like saving the recording the recording metadata and its media to MODB instead of account db.

I believe this code must record same as `cf_record_call`, by calling `kapps_call:start_recording/2`. Lets provide the Pivot test result to managers and then decide what to do next.

icehess (changes requested):

dhuynh-Kazoo:
Put on hold for decision

icehess:
Closing in favor of ticket KPIV-20

PR #806, opened 2024-04-03
[master] KPIV-18: Twiml Dial hangup atrributes cause call crashed

dhuynh-Kazoo:
Due to the incorrect matching type, Kazoo Twiml could not parse the hangup_digit and hangup_dtmf. So the call crashes.

Badmatch issue: https://github.com/2600hz/kazoo-core/blob/e5a7cedef49ac75fce3b223d7e6d983b243bde3c/kazoo_translator/src/converters/kzt_twiml_dial.erl#L365

Missing case for _media_processing_ mode when _record not set and hangupDTMF_ is set:
https://github.com/2600hz/kazoo-core/blob/e5a7cedef49ac75fce3b223d7e6d983b243bde3c/kazoo_translator/src/converters/kzt_twiml_dial.erl#L338


kazoo_translator/src/converters/kzt_twiml_dial.erl
```diff
@@ -332,11 +332,10 @@ request_id(N, Call) -> iolist_to_binary([N, <<"@">>, kapps_call:from_realm(Call)
 
 -spec media_processing(kapps_call:call()) -> kz_term:ne_binary().
 media_processing(Call) ->
-    media_processing(kzt_util:get_record_call(Call), kzt_util:get_hangup_dtmf(Call)).
-
--spec media_processing(boolean(), kz_term:api_binary()) -> kz_term:ne_binary().
-media_processing('false', 'undefined') -> <<"bypass">>;
-media_processing('true', _HangupDTMF) -> <<"process">>.
+    case kzt_util:get_record_call(Call) of
```
dhuynh-Kazoo, line 335: If DID call is recorded, use `process` audio mode, so FreeSWITCH can process DTMF. Otherwise, just use the bypass mode

PR #814, opened 2024-04-11
Upgrade to OTP-26

jamesaimonetti:
- dbg:stop_clear is deprecated
- jiffy was updated to 1.1.1 conversation needed on whether custom flags are still needed


kazoo_schemas/src/kazoo_schemas.app.src
```diff
@@ -1,13 +1,7 @@
...
- ,{registered, [ ]}
- ,{applications, [ kernel
-                 , stdlib
-                 , lager
-                 , kazoo_stdlib
-                 , jesse
-                 ]}
- ]}.
+{application,kazoo_schemas,
+ [{applications,[jesse,kazoo_data,
```
lazedo: can we remove `kazoo_data` from deps ?

PR #816, opened 2024-04-22
KPIV-17: Align TwiML Conference Attributes with Kazoo

dhuynh-Kazoo:


kazoo_translator/src/converters/kzt_twiml_dial.erl
```diff
@@ -75,7 +75,7 @@ exec(Call, [#xmlElement{name='Conference'
     ConfDoc = build_conference_doc(Call, ConfId, ConfProps),
 
     ConfReq = [{<<"Call">>, kapps_call:to_json(Call)}
-              ,{<<"Conference-ID">>, ConfId}
+              ,{<<"Conference-ID">>, 'undefined'} %% hardcode to use ad-hoc conference
```
icehess: Why setting this to `undefined`?

icehess (changes requested):

kazoo_translator/src/converters/kzt_twiml_dial.erl
dhuynh-Kazoo: Hi @icehess, I removed it. Previously in my local test, the conference does not play welcome entry so I considered it as not working. But did the test again and can confirm it worked with ConfId

dhuynh-Kazoo:
The latest commit *4 is for reviewing and will be detached soon, moved to new PR

icehess:
We had discussion about conference features and decided this entry/exit is only support per conference and we expect it is already configured as a profile in the account. This doc should mention this so people know what to do before using this.

PR #818, opened 2024-04-25
[master] KZOO-473 update ledger view

k-anderson:
Update the ledgers view to ignore draft ledgers when computing the balance as well as listing in the UI.  In addition added a view to find just draft ledgers.


k-anderson:
related PRs #819 and #870
supports https://github.com/2600hz/kazoo-iou/pull/37 and https://github.com/2600hz/kazoo-crossbar/pull/581

PR #822, opened 2024-05-07
KPIV-26: fix issues with pivot dial action and updates its offnet_req

icehess:
- Fix DTMF detection in `Dial`, this checks for both if we need to record or detect DTMF and sets to process the media on FS if we need, otherwise bypass the media
- update old code to support new possible values for `record`
- Support new values for `record` to record in mono or stereo (dual)
- Previously `offnet_req` was sent to controller queue which was cf_exec, and was causing offnet_resp there instead of pivot process, this has been changes in [Pivot to acquire queue ](https://github.com/2600hz/kazoo-pivot/pull/29)then use that queue to receive offnet_resp. So now if the Stepswitch find and error and sends resp back (like no resource found error), Pivot can received it instead of just hanging there.
- Code to publish offnet was also updates based on what is in `cf_resource`. Some codes are add as comment and maybe we implement them later if we want.
- Updated old code and bind to other leg when `CHANNEL_BRIDGE` for its events.
- fixed `timeout` is set on offnet causing it to drop call instead of `timeLimit`
- Set DialCallStatus properly for both offnet and conference
- Don't hangup when timeout/timeLimit exceeded or called party hungup, continue render TwiML if there is any.


icehess:
- Master PRs:
    - https://github.com/2600hz/kazoo-pivot/pull/29
    - https://github.com/2600hz/kazoo-core/pull/822
- 5.3 PRs:
    - https://github.com/2600hz/kazoo-pivot/pull/30
    - https://github.com/2600hz/kazoo-core/pull/823

PR #825, opened 2024-05-13
KPIV-17: fix kapps_merged_config:get_ne_binary is not getting the configed value

icehess:
`kapps_merged_config:get_ne_binary` will use `kapps_merged_config:get` which hard coed to get `kz_jspn:get_json_value` ignore the fact that the Key's value maybe not be non-JSON object.

For example `kapps_conference:profile_names/3` uses this function to get profile names but it always returns `undefined` even though the key `profiles.{ProfileName}.base_profile_name` is set! `kapps_merged_config:get` is expecting JSON but it is binary so returns the default value (undefined in this case).


lazedo (approved):

icehess:
5.3 PR: https://github.com/2600hz/kazoo-core/pull/826

PR #834, opened 2024-05-21
[5.3] KPIV-36: Fix issue not playing prompt for TwiML Reject

dhuynh-Kazoo:
Currently, pivot_call receives the stop and executes the hangup command that makes with prompt not work. This case cause error:

https://github.com/2600hz/kazoo-pivot/blob/d19fcc8630cf6cb77a957b1c888ea34a5ca5c65b/src/pivot_call.erl#L240

Call is hangup before so the kapps_call_command:response/4 sent to FS are not valid.


dhuynh-Kazoo:
Related PR: https://github.com/2600hz/kazoo-pivot/pull/32

icehess (approved):

PR #833, opened 2024-05-21
KPIV-35: Processsing length attribute of TwiML Pause

dhuynh-Kazoo:
If the given time is >= 3600, set 3600 seconds.
If the given time is > 0 and< 3600, set that time
If the given time is invalid or not given, set one second

PR #835, opened 2024-05-21
[5.3] KPIV-35: Processsing length attribute of TwiML Pause

dhuynh-Kazoo:
If the given time is >= 3600, set 3600 seconds.
If the given time is > 0 and< 3600, set that time
If the given time is invalid or not given, set one second


icehess:
PRs:
- https://github.com/2600hz/kazoo-core/pull/833
- https://github.com/2600hz/kazoo-core/pull/835

PR #838, opened 2024-05-24
KPIV-30: Using default key for empty finishOnKey attribute

dhuynh-Kazoo:
For the case `<<>> or finishOnKey=""`, kazoo set `undefined`, that seems invalid. We should use default value `#` for this case


kazoo_translator/src/converters/kzt_twiml_util.erl
```diff
@@ -73,7 +73,7 @@ finish_dtmf(Props) -> finish_dtmf(Props, <<"#">>).
 finish_dtmf(Props, Default) when is_list(Props) ->
     case props:get_binary_value('finishOnKey', Props) of
         'undefined' -> Default;
-        <<>> -> 'undefined';
+        <<>> -> Default;
```
icehess, line 76: Returning `undefined` here is intentional. We are suppose to accept empty string here. Empty string will collect dtmf until timeout is reached.

icehess (changes requested):

PR #839, opened 2024-05-28
KPIV-17: add more join conference options to discovery req

icehess:
This updates conference call command to allow more options when a member joins the conference without needing to update code to add new options.

Also updating discovery req to accept more headers so applications send more conference override settings.

Adds accessories for endconf conference flags.

PR #840, opened 2024-05-28
KPIV-17: conference and join member options to dial conference discovery

icehess:
This updates conference discovery req from Dial Conference nouns to pass along some member join options.

Also add support for getting conference id and profile to use from twiml doc.

Developers are expected to use pre-configred conference and profile, wich some conference flags may dyanmically set for the first time when a conference session is started.

Sets join options for member.


icehess:
Requires https://github.com/2600hz/kazoo-core/pull/839 and all of its related linked PRs.

kazoo_amqp/src/api/kapi_conference.erl
```diff
@@ -521,6 +523,8 @@ discovery_req_definition() ->
                 ,{<<"Play-Exit-Tone">>, fun binary_or_boolean/1}
                 ,{<<"Play-Welcome">>, fun kz_term:is_boolean/1}
                 ,{<<"Play-Welcome-Media">>, fun kz_term:is_ne_binary/1}
+                ,{<<"Min-Required-Recording-Participants">>, fun kz_term:is_pos_integer/1}
+                ,{<<"Auto-Record">>, fun kz_term:is_ne_binary/1}
```
jamesaimonetti: I don't think `Auto-Record` is a good name for this property as it suggests, to me, a boolean flag. It seems like the value of this key is a filename for the recording if/when the recording is started.

Is that the purpose of this value, to name the recording?

dhuynh-Kazoo: hi @jamesaimonetti, I updated it with a boolean value. If it is set, we will add the needed fields to the profiles in the conference (https://github.com/2600hz/kazoo-conference/pull/53/commits/00d04b69d08e65219a2f5c60dc8ecdfd5e40a4d5)

kazoo_translator/src/converters/kzt_twiml_dial.erl
```diff
@@ -424,94 +410,72 @@ timelimit_s(Props) ->
...
+                              }
+                             ]),
+
+    ConfReq =
+        props:filter_undefined(
+          [{<<"Call">>, kapps_call:to_json(Call)}
+          ,{<<"Conference-ID">>, Id}
+          ,{<<"Conference-Name">>, Name}
+          ,{<<"End-On-Last-Member-Leave">>, props:is_true('endConferenceOnLastMemberExit', ConfProps, 'undefined')}
+          ,{<<"End-On-Leave">>, props:is_true('endConferenceOnExit', ConfProps, 'undefined')}
```
dhuynh-Kazoo, line 431: hi @icehess,

I have a question regarding the `endConferenceOnExit` which is converted to kazoo conference flag `End-On-Leave` (`endconf` in ecallmgr).

As the twilio doc: 

> If a participant has this attribute set to true, then when that participant leaves, the conference ends and all other participants drop out. This defaults to false. This is useful for implementing moderated conferences that bridge two calls and allow either call leg to continue executing TwiML if the other hangs up.

But from the FS doc, endconf of member-flags means

> Ends conference when all members with this flag leave the conference after profile param endconf-grace-time has expired

My testing with 3 participants: 1 member and 2 moderators (End-On-Leave=true). One moderator leaves, the conference continued. Is that the case? It seems incorrect with the Twilio doc. Should we use `End-On-Last-Member-Leave` ~ `mandatory_member_endconf` instead?

icehess: `mandatory_member_endconf` from fs side (`End-On-Last-Member-Leave` from kazoo side) is actually what ends the conference after all members with that flag exit. `endconf` will end the conference for all when a member with that flag exits.

dhuynh-Kazoo: Thank you @icehess, I just found the comment from the FS source. In the mod_conference doc some points are not correct.

```
		/* End conference when any member with "endconf" flag disconnects */
		if (conference_utils_member_test_flag(member, MFLAG_ENDCONF)) {
			conference_utils_set_flag_locked(conference, CFLAG_DESTRUCT);
		}

		/* End conference only if all mandatory members have disconnected */
		if (conference_utils_member_test_flag(member, MFLAG_MANDATORY_MEMBER_ENDCONF)) {
			if (!--conference->end_count) {
				conference->endconference_time = switch_epoch_time_now(NULL);
			}
		}
```

icehess:
PRs:
- master:
    - https://github.com/2600hz/kazoo-core/pull/840
    - https://github.com/2600hz/kazoo-core/pull/839
    - https://github.com/2600hz/kazoo-crossbar/pull/574
    - https://github.com/2600hz/kazoo-conference/pull/50
    - https://github.com/2600hz/kazoo-ecallmgr/pull/246
    - https://github.com/2600hz/kazoo-ecallmgr/pull/260
- 5.4:
    -  https://github.com/2600hz/kazoo-core/pull/978
    - https://github.com/2600hz/kazoo-core/pull/924
    - https://github.com/2600hz/kazoo-crossbar/pull/595
    - https://github.com/2600hz/kazoo-conference/pull/51
    - https://github.com/2600hz/kazoo-ecallmgr/pull/253
    - https://github.com/2600hz/kazoo-ecallmgr/pull/263

PR #844, opened 2024-05-29
[5.3] KPIV-26: clean xmText when filtering empty text in XML

dhuynh-Kazoo:
Fix filter_empty_text is not properly cleaning xmlText values when filtering

Fix logging issue from https://github.com/2600hz/kazoo-core/pull/822


dhuynh-Kazoo:
Original PR: https://github.com/2600hz/kazoo-core/pull/822

PR #846, opened 2024-05-29
KPIV-36-II: Returning stop and stop_call depends on TwiML verbs

dhuynh-Kazoo:
Return _stop and stop_call_ for types of TwiML verbs:

- `stop_call` is used to hard stop the call with TwiML Reject, Hangup,...
- `stop` for other verbs


icehess:
Closing in favor of https://github.com/2600hz/kazoo-core/pull/920 and https://github.com/2600hz/kazoo-core/pull/920

PR #848, opened 2024-05-30
KPIV-20: adding silence variables to record_call

icehess:
Adds support for stop recording on silence detection for freeswitch `record_session` app. Silence detection should be set explicitly to get enabled.


icehess:
Related PRs:
- 5.4:
    - https://github.com/2600hz/kazoo-ecallmgr/pull/252
    - https://github.com/2600hz/kazoo-core/pull/869
- master
    - https://github.com/2600hz/kazoo-ecallmgr/pull/247
    - https://github.com/2600hz/kazoo-core/pull/848

lazedo (commented):
would prefer a `Record-Silence-Detection` object with its related variables there

lazedo (changes requested):
`defaults` and `checks` should be done here

lazedo (approved):

PR #854, opened 2024-06-03
[5.4] KPIV-39: sanity check the play, say and gather attribute values (#837)

dhuynh-Kazoo:
This checks and sanitize the attribute values for some verbs and make sure the acll won't be crashed on invalid values. For Say and Play, get account settings for engine, lang if possible.

Fixes Gather with infinity numDigits crash.


icehess:
5.4: https://github.com/2600hz/kazoo-core/pull/853
master: https://github.com/2600hz/kazoo-core/pull/837

PR #855, opened 2024-06-03
schema properties must specified their type

icehess:


icehess:
requires: https://github.com/2600hz/kazoo-crossbar/pull/576

PR #857, opened 2024-06-06
KPIV-20: use record_call command in kzt_twiml for call recording

icehess:
- Using kapps_call:start_recording/2 for Twiml Record to let media_mgr handle storing stuff
- Using kapps_call:start_recording/2 for Twiml Dial when enabling Record attribute to let media_mgr handle storing stuff
- Modify the wait_for_record event logic
- Modify the wait_for_offnet_evets with and without Record attribute

PR #859, opened 2024-06-07
KPIV-20: Testing new implementation for TwiML Record verb

dhuynh-Kazoo:
- Continue loop if terminator digits not correct
- Using default PUT method to put recorded file to server


icehess:
Thanks, cherry picked to https://github.com/2600hz/kazoo-core/pull/857

PR #867, opened 2024-06-12
KCRO-206 : As a developer, I want an API for crud operations on a list of from addresses so that can DKIM to sign emails

manushiperera:
Previously, DKIM information was common for the account and unable to configure based on the available email addresses of the account. Now, a new schema is defined to save the email address and related DKIM information.

A new email template is introduced to send the pin code for the email verification purposes and generated new AMQP notification event to teletype for email verification


manushiperera:
Related PRs
https://github.com/2600hz/kazoo-teletype/pull/73
https://github.com/2600hz/kazoo-properly/pull/208
https://github.com/2600hz/kazoo-crossbar/pull/580

PR #868, opened 2024-06-13
[5.3] KPIV-20: adding silence variables to record_call

icehess:
Adds support for stop recording on silence detection for freeswitch `record_session` app. Silence detection should be set explicitly to get enabled.

* adding silence variables to record_call

Adds support for stop recording on silence detection for freeswitch `record_session` app. Silence detection should be activated explicitly.

* move to its own object

* check values in publishing side

* move timeout validation to kapi

* intial timeout is for when no speech deteced, final for after first speech detected

(cherry picked from commit 118f6ab88505f690518cb13bec1aad758962bde9)


icehess:
Closing, this is going to be on next version of 5.4

PR #880, opened 2024-06-20
[5.4] KPIV-43: TwiML Dial Record attribute not working

dhuynh-Kazoo:
From this merged PR [kazoo-core: //github KPIV-26: fix issues with pivot dial action and updates its offnet_req (#822) - Fix DTMF detection in](https://github.com/2600hz/kazoo-core/commit/53947e61d12d4c7a3e3921d55c917e7b78b400a3)

Not able to record a call when recording is enabled. Call crashes after getting bridged.

Test case for TwiML Dial:

```
<?php
header("content-type: text/xml");
echo "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n"
?>

<Response>
    <Dial record="true">
        +14157365123
    </Dial>
</Response>
```


icehess:
Closing this is as it is going to be addressed by https://github.com/2600hz/kazoo-core/pull/857

PR #890, opened 2024-07-12
Send display 5.3

kusalfernando:


kazoo_amqp/priv/couchdb/schemas/kapi.dialplan.command.json
```diff
@@ -90,7 +90,8 @@
                 "tones",
                 "transfer",
                 "tts",
-                "unbridge"
+                "unbridge",
+                "send_display"
```
danielfinke, line 94: add this alphabetically in the list (just before `send_dtmf`)
also recommend you move the code in `kapi_dialplan` around, above `send_dtmf`, too

danielfinke (changes requested):

PR #893, opened 2024-07-17
KZOO-498: Validate DUO Universal codes for MFA

jamesaimonetti:


jamesaimonetti:
Helpful PR: https://github.com/2600hz/kazoo-crossbar/pull/587
Local testing: https://github.com/2600hz/kazoo-properly/pull/211

PR #896, opened 2024-07-18
KZOO-499: better bulk del of tombstoned docs

danielfinke:
When bulk deleting docs using BigCouch 0.4.2, previously-deleted docs with tombstones were able to be repeatedly deleted, each time increasing their revision value. In CouchDB 2+, deleting a tombstoned doc will result in a conflict. This change ensures that, during bulk rev fetch, tombstoned docs are not included in the list of docs to be deleted. This matches the behaviour of deleting docs that never existed or for which there is no tombstone at all.

Fixes a 500 error that occurs when `knm_phone_number` tries to delete a number from the previously-assigned-to account DB when the number has already been deleted.

I'm not sure that the semantics of `del_doc` vs `del_docs` make sense though. This is the matrix of behaviours:
Function | `Doc` argument | Doc exists | Has tombstone | Result
-------- | -------------- | ---------- | -------------- | ------
`del_doc` | `<<"+12192550447">>` | `false` | `false` | `{error,not_found}`
`del_doc` | `<<"+12192550447">>` | `true` | `false` | `{ok,<DELETED_DOC>}`
`del_doc` | `<<"+12192550447">>` | `true` | `true` | `{error,not_found}`
`del_docs` | `[<<"+12192550447">>]` | `false` | `false` | `{ok,[]}`
`del_docs` | `[<<"+12192550447">>]` | `true` | `false` | `{ok,[<DELETED_DOC>]}`
`del_docs` | `[<<"+12192550447">>]` | `true` | `true` | **before PR**: conflict, **after PR**: `{ok,[]}`
--- | --- | --- | --- | ---
`del_doc` | `kz_doc:set_id(kz_doc:new(), <<"+12192550447">>)` | `false` | `false` | creates new tombstoned doc
`del_doc` | loaded doc | `true` | `false` | `{ok,<DELETED_DOC>}`
`del_doc` | forged doc with latest rev | `true` | `true` | conflict
`del_doc` | forged doc with no rev | `true` | `true` | conflict
`del_docs` | [forged doc with no rev] | `false` | `false` | `{ok,[]}`
`del_docs` | [loaded doc] | `true` | `false` | `{ok,[<DELETED_DOC>]}`
`del_docs` | [forged doc with latest rev] | `true` | `true` | conflict
`del_docs` | [forged doc with no rev] | `true` | `true` | `{ok,[]}`

so the inconsistent parts are:
1. `del_docs` considers it OK (`{ok,[]}`) to delete non-existent or tombstoned docs, `del_doc` gives an `{error,not_found}`
2. `del_doc` creates tombstones when deleting docs that don't already exist

After PR: `del_docs([<<"+12192550447">>])` is equal of `del_docs([forged doc with no rev])` since in both cases there should be no known rev


danielfinke:
OK'd by @jamesaimonetti in [Slack](https://2600hz.slack.com/archives/CH9JJLZED/p1721328423216599)

PR #904, opened 2024-07-23
[5.3] KZOO-502: AMQP message handling for CRM call interactions

kusalfernando:
Include CRM call interaction header in caller info response definition.


kazoo_amqp/priv/couchdb/schemas/kapi.crm.caller_info_reply.json
```diff
@@ -15,6 +15,9 @@
             "enum": [
                 "caller_info_reply"
             ]
+        },
+        "Query-Interaction-Response": {
+            "type": "string"
```
danielfinke: It looks like the response type is object? https://github.com/2600hz/kazoo-crm/pull/95/files#diff-2bcb8628ac724d956f301c55bc810c8cb3ba33a118e44de3d22d8a9560c0c7acR128
```suggestion
            "type": "object"
```

danielfinke (approved):

danielfinke:
@kusalfernando branch needs to be rebased

PR #908, opened 2024-07-26
KZOO-456: track proc mailbox/heap growth, use `off_heap` messages in `gen_listener`

danielfinke:
`kz_monitor`: use `erlang:system_monitor/2`
- track GC time, total heap size, and MQ length
- MQ length warnings are only logged as they exceed the high threshold
- long GC/large heap are logged each time event is received
  - but throttle logging for busy monitored processes to minimize load on `kz_monitor` itself
- add refs to discussions/blogs about `{message_queue_data, off_heap}`
- general tidying (types, conventions)

remove `gen_listener` GC polling:
- remove `auto_gc` start param/`maybe_gc/1` fun
- by default, monitor all `gen_listener`s using `kz_monitor`
- enable `off_heap` `message_queue_data` flag for all `gen_listener`s
- refactor `start_link`/`enter_loop`/fix their types
  - init funs too, to avoid position-based type assumptions/collapse to `term()`
- types and doc blocks

`kz_monitor` perf impact on idle Kazoo VM:
- 36k-40k reductions per 5 seconds
- 7k-10k reductions per 5 seconds when polling is commented out
  - suggests the ~30k delta will be gone when we move to OTP 27 impl.

## Related PRs
- https://github.com/2600hz/kazoo5/pull/162
- https://github.com/2600hz/kazoo-ecallmgr/pull/250


lazedo:
the only thing I would change is to move kz_monitor out of kazoo_apps because with this change kazoo_amqp will depend on kazoo_apps and apps cannot start kazoo_amqp without starting kazoo_apps (which starts a lot of other stuff like kazoo_data). maybe create a new kazoo_monitor app or add it to kazoo_stdlib if we intend to reuse for other procs other than gen_listener ones, or into kazoo_amqp if gen_listener is ok for now.

lazedo (approved):

PR #917, opened 2024-07-30
KZOO-456: `kz_monitor` per-proc thresholds

danielfinke:
Allow tracked procs to specify custom thresholds for GC time, large heap size, and low/high message queue len watermarks. Can use `kz_monitor:set_thresholds/2` to update thresholds during runtime. Warnings are suppressed when custom thresholds are not met.

Set heap size threshold to 100,000 words for `kz_nodes_listener`.

See: https://2600hz.slack.com/archives/CH9JJLZED/p1722339441962129


kazoo_amqp/src/gen_listener.erl
```diff
@@ -557,11 +557,15 @@ init(Module, Params, ModuleState, TimeoutRef) ->
 maybe_monitor_me(Params) ->
     case props:is_true('kz_monitor', Params, ?DEFAULT_KZ_MONITOR) of
```
lazedo: props:get_value/3 ?

lazedo (commented):
would prefer to use a map with `Options` where `heap_size` is one of the options, but I'm ok for now to use this format.

lazedo (approved):

PR #920, opened 2024-07-31
KPIV-46: update reject code and reason for Reject, use queque hangup

dhuynh-Kazoo:
Using the correct code `603` for `CALL_REJECTED`. Also fix issues when reject wants to play a media and the call may stay up and play the media in the loop. Pivot now uses queue hangup to allow the queue to be cleared before going down.


dhuynh-Kazoo:
hi @jamesaimonetti, could you please check that? I am not sure why the current code returns 503 for **rejected** reason. Would you happen to know the reason for that? From the FS doc, I think it should be the 603 code

sanjeevie6353:
@dhuynh-Kazoo can you please add a bit more context to the description on what you are going to do here. Just curious to know more.

These are the 2 SIP response code descriptions that I found on the 2 mentioned codes. 

**603: decline**
A callee’s end system was successfully contacted but the user does not wish to or cannot participate.

**503: service unavailable**
Temporary overloading or maintenance of a server means it is currently unable to process the request.

dhuynh-Kazoo:
hi @sanjeevie6353, the code is related to kazoo-pivot application, when using TwiML Reject verb to reject an incoming call with a **rejected** reason, the current code returns 503 code. You can see the call in SmartPBX call logs with reason NORMAL_TEMPORARY_FAILURE.

Following the FreeSwitch document, 503 is NORMAL_TEMPORARY_FAILURE and 603 is CALL_REJECTED

dhuynh-Kazoo:
Related PR: https://github.com/2600hz/kazoo-pivot/pull/40

icehess (approved):

icehess:
Related PRs:
- master:
    - https://github.com/2600hz/kazoo-core/pull/920
    - https://github.com/2600hz/kazoo-pivot/pull/40
- 5.4:
    - https://github.com/2600hz/kazoo-core/pull/923
    - https://github.com/2600hz/kazoo-pivot/pull/42
- 5.3:
    - https://github.com/2600hz/kazoo-core/pull/926
    - https://github.com/2600hz/kazoo-pivot/pull/44

PR #921, opened 2024-07-31
KPIV-38: properly set authorizing id and loopback bowout when creating fwd in twiml

icehess:
When using multiple Number nouns, we create a temp device endpoint with unconditional call forwarding. It doesn't set authorizing-id, which ecallmgr then tries to authz the loopback channel for both channel create event and fetch dialplan events. dialplan event has the number prefixed with `sip:` which will then cause jonny unable to find the billing info and not unauthorized the channel, the call will killed after that.

This adds authorizing-id to the bridge command so ecallmgr will now it the channels are authorized.

We use the authorizing-id from existing call if there is any, otherwise we use account as endpoint.

This also fixes another issue in multi Number case, since we use call forwarding, loopback channel events will be sent and after the bridge they will bowout and callflow kills the call. This adds `Loopback-Bowout` to avoid sending these events from FS when loopback channels are destroyed.


icehess:
Related PRs:
- master:
    - https://github.com/2600hz/kazoo-core/pull/921
    - https://github.com/2600hz/kazoo-pivot/pull/41
- 5.4:
    - https://github.com/2600hz/kazoo-core/pull/925
    - https://github.com/2600hz/kazoo-pivot/pull/43
- 5.3:
    - https://github.com/2600hz/kazoo-core/pull/927
    - https://github.com/2600hz/kazoo-pivot/pull/45

icehess:
Update to fix bowout must be false https://github.com/2600hz/kazoo-core/pull/921

PR #923, opened 2024-08-02
[5.4] KPIV-46: update reject code and reason for Reject, use queque hangup

dhuynh-Kazoo:
Using the correct code `603` for `CALL_REJECTED`. Also, fix issues when the reject wants to play a media, and the call may stay up and play the media in the loop. Pivot now uses queue hangup to allow the queue to be cleared before going down.


icehess:
Related PRs:
- master:
    - https://github.com/2600hz/kazoo-core/pull/920
    - https://github.com/2600hz/kazoo-pivot/pull/40
- 5.4:
    - https://github.com/2600hz/kazoo-core/pull/923
    - https://github.com/2600hz/kazoo-pivot/pull/42
- 5.3:
    - https://github.com/2600hz/kazoo-core/pull/926
    - https://github.com/2600hz/kazoo-pivot/pull/44

PR #926, opened 2024-08-02
[5.3] KPIV-46: update reject code and reason for Reject, use queue hangup 

dhuynh-Kazoo:
…(#920)

Using the correct code `603` for `CALL_REJECTED`. Also fix issues when reject wants to play a media and the call may stay up and play the media in the loop. Pivot now uses queue hangup to allow the queue to be cleared before going down.

* Using 603 code for rejected reason
* Using {'stop', Call} for execution completed
* remove unsed type

---------


dhuynh-Kazoo:
Origin PR: https://github.com/2600hz/kazoo-core/pull/920

PR #927, opened 2024-08-02
[5.3] KPIV-38: properly set authorizing id and loopback bowout when creatin…

dhuynh-Kazoo:
…g fwd in twiml (#921)

When using multiple Number nouns, we create a temp device endpoint with unconditional call forwarding. It doesn't set authorizing-id, which ecallmgr then tries to authz the loopback channel for both channel create event and fetch dialplan events. dialplan event has the number prefixed with `sip:` which will then cause jonny unable to find the billing info and not unauthorized the channel, the call will killed after that.

This adds authorizing-id to the bridge command so ecallmgr will now it the channels are authorized.

We use the authorizing-id from existing call if there is any, otherwise we use account as endpoint.

This also fixes another issue in multi Number case, since we use call forwarding, loopback channel events will be sent and after the bridge they will bowout and callflow kills the call. This adds `Loopback-Bowout` to avoid sending these events from FS when loopback channels are destroyed.


dhuynh-Kazoo:
Origin PR: https://github.com/2600hz/kazoo-core/pull/921

PR #930, opened 2024-08-05
KZOO-509: Set common SSL options

jamesaimonetti:
Create a module for keeping common SSL options in one place. In OTP 26+ it is necessary to define the `verify` parameter with either `verify_none` or `verify_peer` and include what certificates should be used.


jamesaimonetti:
Closes #922 in favor of PR with JIRA ticket

icehess (approved):

PR #935, opened 2024-08-09
KZOO-503: Validate hostnames for SIP using SRV

jamesaimonetti:
Previously, only A records were consulted for whether a hostname resolved appropriately.

Now, check the SRV records first, falling back to previous A record checks if no SRV are configured.

Also introduce a callback for jesse for doing format validation


jamesaimonetti:
Testing: https://github.com/2600hz/kazoo-properly/pull/213
Crossbar: https://github.com/2600hz/kazoo-crossbar/pull/597
Jesse bump: https://github.com/2600hz/kazoo5/pull/163

icehess (approved):

PR #948, opened 2024-09-02
KZOO-526 Delete attachment does not work if attachment is saved in aw…

egor-ezhov-ooma:
## Solution text

Author: Egor Ezhov

Bug ID: KZOO-526 Delete attachment does not work if attachment is saved in aws s3

Details:

In the initial implementation (https://github.com/2600hz/kazoo-core/pull/372) the attachment is deleted based on the handler from the storage doc. It needs to be changed to use S3 data from the media doc which contains all required data.

The solution proposes:
- using S3 data from the media doc to delete attachment
- using attachment options to call `kzs_plan:plan` from the `kz_datamgr:delete_attachment` (same is used in the `put_attachment`, `fetch_attachment` and `update_attachment_metadata`) to correctly identify doc with media
- removing attachment's data from the `pvt_attachments` section in the media doc if the attachment is deleted. Data from the `_attachments` is deleted automatically by CouchDB. 
### Test Cases performed

1. Configured storage plan with AWS S3 for user data
2. Upload user's photo
3. Check that user's photo is stored in the AWS S3
4. Delete user's photo
5. Check that user's photo is deleted from the AWS S3.
6. Check that attachment's info is removed from the user doc


lazedo:
this is not how it works @egor-ezhov-ooma . `data plan` is for access the document that contains the attachment, the `handler` from `pvt_attachment` already has the information it needs to delete the attachment

lazedo:
the `update` to the doc with the attachment removed  is good, but needs to check if its needed. this was missing from initial contribution https://github.com/2600hz/kazoo-core/pull/372

egor-ezhov-ooma:
@lazedo , yes, you are right. I checked it deeper, looks like we did not check 'delete_enabled' in the S3 data which is stored in the doc. I will rework the solution. Anyway, the plan expects the doc type, not the doc id. I think it should be fixed also.

lazedo (changes requested):
the `BPC` already contains the merged settings for deleting the attachment, seems like `s3` changes are not required as it looks like a different way of doing the same thing. I think the only change that applies here is the removal of `attachment`
 from doc when its info is stored in pvt_attachments.

egor-ezhov-ooma:
@lazedo S3 var is required/used only to check the param should we remove the real attachment from the AWS S3 in this line
```
case kz_att_util:should_del_att(decode_retrieval(S3)) of
```

if S3 param has this flag, then `BPC` is used

kazoo_data/src/kz_datamgr.erl
```diff
@@ -1061,16 +1061,47 @@ put_attachment(DbName, DocId, AName, Contents, Options) ->
...
+    {'ok', <<>>};
+handle_delete_attachment({'ok', _} = Resp, DbName, DocId, AName) ->
+    case open_cache_doc(DbName, DocId) of
+        {'ok', Doc} ->
+            kzs_cache:flush_cache_doc(DbName, DocId),
+            UpdatedDoc = kz_doc:delete_attachment(Doc, AName),
+            _ = save_doc(DbName, UpdatedDoc);
+        {'error', _}=E -> E
+    end,
+    Resp;
```
lazedo: no need when attachment is on couch

lazedo (changes requested):

kazoo_data/src/kz_datamgr.erl
egor-ezhov-ooma: it will do nothing if attachment in the couch, since `_attachment` section is already deleted and `pvt_attachment` is not set in this case. I don't think that additional validation of handler (i.e. where the attachment is stored) is required.

lazedo: you are adding a new revision to the doc when it's not necessary besides the http call to couch and couch internal things that will be triggered.
you don't need to validate the handler itself, what you need is to save only if the attachment was removed from `pvt_attachments` , ie, the ones from an external handler.

unnecessary calls/updates should be avoided when known

egor-ezhov-ooma: updated

jamesaimonetti:
@egor-ezhov-ooma i rebased your branch against latest master and force pushed to here. Should also kick off CircleCI checks (will have to see how to get you added to circle to run CI)

jamesaimonetti:
@egor-ezhov-ooma added two commits - one for unhandled returns and one to fix edoc tooling and a formatting change.

enoskov-ooma:
@jamesaimonetti circleCI doesn't work for new users. It doesn't work for me either or works time-to-time. Tried to address it with Sean Wysor but vainly. A workaround is to ask somebody with old account to restart.

PR #956, opened 2024-09-12
rebar3: fix release creation in default rebar3 config

icehess:
rebar3 will create `.app` file for any app or deps if it is missing and assumes. In case the app doesn't have `start/0` this will cause erlang fails to boot up from a release which tries to start the app on boot.

`qrcode` doesn't have `.app` file and doesn't have `qrcode:start/0` so a rebar3 release fails to boot and crashes. Since we only need qrcode to be loaded and not needed to be started, moving it to `include_applications` which only ensures the app is loaded.

Per documentation:

> `included_applications` - All applications included by this application. When this application
> is started, all included applications are loaded automatically, but
> not started, by the application controller. It is assumed that the
> top-most supervisor of the included application is started by a
> supervisor of this application.


lazedo:
not seeing this behaviour @icehess, maybe its related to other changes ?

icehess:
@lazedo not sure what other changes. My release always crashes during boot. If I use `start_clean` boot instead and manually start kazoo_apps I see it is crashing when trying to call `qrcode:start`.

And i know that rebar is creating qcode.app file and sets `{mod,{qrcode,[]}}`. I think you are not seeeing it beacuse you hard coded qrcode to just be in the path.

From you diff of `rebar.config.script`:

```diff
+                     Deps = [{qrcode, none}
+                            ,{reloader, none}
+                            ],
```

icehess:
```
23:05:21.617 [error] |0000000000|qrcode:Undefined(<0.1424.0>) CRASH REPORT Process <0.1424.0> with 0 neighbours exited with reason: call to undefined function qrcode:start(normal, []) in application_master:init/4 line 142
23:05:21.617 [error] |00000000000|kapps_controller:119(<0.471.0>) ecallmgr_extension could not start: {qrcode,{bad_return,{{qrcode,start,[normal,[]]},{'EXIT',{undef,[{qrcode,start,[normal,[]],[]},{application_master,start_it_old,4,[{file,"application_master.erl"},{line,293}]}]}}}}}
23:05:21.632 [error] |0000000000|qrcode:Undefined(<0.1427.0>) CRASH REPORT Process <0.1427.0> with 0 neighbours exited with reason: call to undefined function qrcode:start(normal, []) in application_master:init/4 line 142
23:05:21.632 [error] |00000000000|kapps_controller:119(<0.471.0>) jonny5 could not start: {qrcode,{bad_return,{{qrcode,start,[normal,[]]},{'EXIT',{undef,[{qrcode,start,[normal,[]],[]},{application_master,start_it_old,4,[{file,"application_master.erl"},{line,293}]}]}}}}}
```

icehess:
Hold on merging this. I think the proper fix is to move `qrcode.app` from its `ebin` to `src/qrcode.app.src` so rebar won't manipulate it.
Need ask @jamesaimonetti to do it since I don't have push access to erlang-qrcode.

lazedo:
I think my diff is doing the right thing :)

icehess:
@lazedo Curios, can you check your qrcode `.app` in ebin folder, does it have `mod` line. Also check your release `.rel` file, does it have `qrcode` without `Type` or the type is set `none`?

Even with type set to `none` any kapps that depends on kazoo_auth directly or indirectly, will ensure all start it which will also starts qrcode and that is where i get that above error like for jonny5 which depends on kazoo_call which depends on kazoo_auth.

So what is happening, rebar modifies the ebin `.app` file and adds `mod` line. In our relx, we only add kazoo apps and core to the release app lists and set their type to `load`. Rebar then also adds dependencies of all those apps we want in the release to the release. But it doesn't set the type as we do, so no type means it is permanent and erlang rel will load and start the app during boot up.

So even if you add `{qrcode, none}` it just won't load and start during _boot_. but when jonnyy or any app tries to start kazo_auth it still starts `qrcode` and i expect you get that error above.


Anyway all this shenanigan can be saved by just move the `.app` file to src folder as `.app.src` so rebar won't be stupid and change it.

lazedo:
```
{application,kazoo_auth,
             [{applications,[base64url,crypto,kazoo_amqp,kazoo_caches,
                             kazoo_documents,kazoo_stdlib,kazoo_web,kernel,
                             lager,public_key,qrcode,stdlib]},
              {description,"Kazoo Authorization Library"},
              {vsn,"4.0.0"},
              {modules,[kazoo_auth_app,kazoo_auth_maintenance,kazoo_auth_sup,
                        kz_auth,kz_auth_apps,kz_auth_bearer,kz_auth_cert,
                        kz_auth_claims,kz_auth_client,kz_auth_identity,
                        kz_auth_init,kz_auth_jwt,kz_auth_keys,
                        kz_auth_listener,kz_auth_otp,kz_auth_password,
                        kz_auth_profile,kz_auth_providers,kz_auth_qrcode,
                        kz_auth_scope,kz_auth_token_util,kz_auth_util,
                        kz_mfa_auth,kz_mfa_duo,kz_mfa_duo_universal,
                        kz_mfa_otp]},
              {registered,[kz_auth_rsa,kazoo_auth_public_keys_cache,
                           kazoo_auth_tokens_cache,kazoo_auth_app,
                           kazoo_auth_sup]},
              {mod,{kazoo_auth_app,[]}}]}.
```

```
{application,qrcode,
             [{applications,[crypto,kernel,stdlib]},
              {description,"QRCode Encoder"},
              {vsn,"1.0.3"},
              {modules,[base32,bits,gf256,qrcode,qrcode_demo,qrcode_mask,
                        qrcode_matrix,qrcode_png,qrcode_reedsolomon]},
              {mod,{qrcode,[]}},
              {registered,[]},
              {env,[]}]}.
```

lazedo:
I think there may be some mixed fixed things in erlang-qrcode repo, branches 2600hz and 2600hz-otp-26
I'm using 650d9bc0d3d052cb5689a4b4fda161eb8d72a888 from 2600hz branch and it works for me. there is a later commit on 2600hz but there are other on 2600hz-otp-26

so maybe fix the erlang-qrcode repo instead of here, otherwise you would have to go thru all apps and do the same for all deps that are libraries

icehess:
Fixed in https://github.com/2600hz/erlang-qrcode/commit/cbb74b45d513dddae34e40740d399d04c1946ba0

PR #963, opened 2024-09-19
KPIV-38: handle undefined when calculating presence id

icehess:
When creating an endpoint (in this case a call forward endpoint), if the endpoint is an account doc, calculating presence id will crash beacuse the endpoint doc doesn't have the required private keys (pvt_account_id) and realm is undefined.

Kazoo translator will try to create a call forwarding endpoint using authorizing ID if there is any, if not it will try to use the account doc as endpoint when creating call forwarding endpoint.


kazoo_endpoint/src/kz_endpoint_v5.erl
```diff
@@ -406,7 +406,9 @@ merge_attributes(Device, <<"device">>, Options, Keys) ->
     Endpoint = kz_json:set_value(<<"owner_id">>, kz_doc:id(Owner), Device),
     merge_attributes(Keys, Options, endpoint_account(kz_doc:account_id(Device), Options), Endpoint, Owner);
 merge_attributes(AccountDoc, <<"account">>, Options, Keys) ->
-    merge_attributes(Keys, Options, AccountDoc, kz_json:new(), kz_json:new());
+    kz_json:values(kz_doc:private_fields(AccountDoc)
+                  ,merge_attributes(Keys, Options, AccountDoc, kz_json:new(), kz_json:new())
+                  );
```
lazedo, line 411: what's this @icehess ? its returning `empty` endpoint with type and id, please revert

PR #974, opened 2024-09-30
KPIV-17: Add Wait-For-Moderator in Discovery Request 

dhuynh-Kazoo:
Add `Wait-For-Moderator` in the Discovery Request object if we use Twiml Dial Conference with `startConferenceonEnter`.

if we have `startConferenceonEnter=true` so` #kapps_conference{wait_for_moderator=true}`. Otherwise, the default is false. the `wait_for_moderator` already implemented but not defined in DiscoveryReq so that attribute does not work!!!


dhuynh-Kazoo:
hi @icehess, We missed the Wait-For-Moderator in DiscoveryReq API Object

icehess:
Are you using testing with the correct branch? It is already set in the original PR:
https://github.com/2600hz/kazoo-core/pull/840/files#diff-8a9be57633e146df21809664161424bdfe8588ac76c05d07a2552c79b69ca9e1

PR #980, opened 2024-10-11
KECA-41: add stream command to control streaming call media to websocket

icehess:
Adding dialplan command for start/stop/pause/resume kz_audio_fork and streaming call's media to a WebSocket.

The stream will the media as raw audio with PCMU (mu-law) encoding. Sample rate and number of channels are configurable.

The stream can be stopped, pause or resume. By default if no Stream-ID is given, these commands will act on all attached audio forks.


icehess:
PRs:
   - master:
       - https://github.com/2600hz/kazoo-core/pull/980
       - https://github.com/2600hz/kazoo-core/pull/1023
       - https://github.com/2600hz/kazoo-ecallmgr/pull/265
       - https://github.com/2600hz/kazoo-ecallmgr/pull/269
   - 5.4:
       - https://github.com/2600hz/kazoo-core/pull/1036
       - https://github.com/2600hz/kazoo-ecallmgr/pull/270

kazoo_call/src/kapps_call_command.erl
```diff
@@ -1909,6 +1915,86 @@ b_record_call(MediaName, Action, TimeLimit, Terminators, Call) ->
...
+%% @doc Produces the low level AMQP request for stream command to control streaming
+%% the call's media to a websocket server.
+%% @end
+%%------------------------------------------------------------------------------
+-spec stream(kz_term:ne_binary(), kapps_call:call()) -> 'ok'.
+stream(Action, Call) ->
+    stream(Action, [], Call).
+
+-spec stream(kz_term:ne_binary(), kz_term:proplist(), kapps_call:call()) -> 'ok'.
+stream(Action, Options, Call) ->
```
lazedo: change to `stream_command` to follow other commands

lazedo (changes requested):
add `block` for start_stream like other commands. `b_start_stream` that sends and waits for return

I would also like to see stream_`action` instead of `action`_stream, `stream_start` instead of `stop_stream`

kazoo_call/src/kapps_call_command.erl
```diff
@@ -1909,6 +1918,150 @@ b_record_call(MediaName, Action, TimeLimit, Terminators, Call) ->
...
+%%------------------------------------------------------------------------------
+-spec stream_command(kz_term:ne_binary(), kapps_call:call()) -> 'ok'.
+stream_command(Action, Call) ->
+    stream_command(Action, [], Call).
+
+-spec stream_command(kz_term:ne_binary(), kz_term:proplist(), kapps_call:call()) -> 'ok'.
+stream_command(Action, Options, Call) ->
+    Command = [{<<"Application-Name">>, <<"stream">>}
+              ,{<<"Action">>, Action}
+              ,{<<"Insert-At">>, <<"now">>}
```
lazedo: commands that expect a return should not set `{<<"Insert-At">>, <<"now">>}`

lazedo (approved):

PR #988, opened 2024-10-21
Document kz_app_config

jamesaimonetti:


kazoo_stdlib/src/kz_app_config.erl
```diff
@@ -33,6 +35,17 @@
...
+-spec init() -> 'ok'.
+init() ->
+    Terms = [{'kz_account_config', fun kapps_account_config:get_global/3}
+            ,{'kz_config', fun kapps_config:get/2}
+            ],
+    lists:foreach(fun set_config/1, Terms).
+
+set_config({Key, Value}) ->
+    persistent_term:put(Key, Value).
+
```
lazedo: why move this here ? can we leave this in kazoo_apps_init ? that creates a unnecessary reference/dependency to kapps_account_config and kapps_config and dependencies was one of the reasons for creating the module.

lazedo: I think you can leave the `set_config` and ensure proper `Key` is used

jamesaimonetti: it was the leaking of persistent term keys outside the module that i was concerned about.

lazedo: or you can add explicit funs set_config/1 and set_account_config/1 that receive a fun/2 and fun/3 and set kz_config and kz_account_config, these last could be transformed to a macro

lazedo: what we don't want is to create a reference/dependency

jamesaimonetti: noted, will make that change

lazedo (approved):

PR #990, opened 2024-10-24
KPIV-47: implement <Stream> TwiML to send call audio to a websocket

dhuynh-Kazoo:
This adds ability for Pivot TwiML translator to streams the call media using Pivot as proxy between our media server and remote developer controlled websocket server.

Translator sends the processed data from Stream attributes and let pivot handle the rest.

Pivot will issue the stream commands to ecallmgr and uses itself to act as proxy between freeswitch and remote websocket. The proxy process will send appropriate message to remote websocket url base on the type of stream.

Communications between pivot process, response process and proxy process is through passing along their respective PID and use normal Erlang message passing. Currently remote node RPC is not support and we use hostname of the calling pivot process nodes when creating the URL for freeswitch to connect to.

Pivot issues the stream start to ecallmgr, which goes to freeswitch.

Freeswitch will connect to pivot on the same host that issued the start command, and sends the playload the pivot passed along from start command. Proxy process setups websocket between itself and freeswitch and waits for freeswitch to send this payload.

When the payload received, the proxy will use this info to get the Call (kapps_call:call()) and the Stream-ID (the result of stream start dialplan command) from calling pivot process then tries to connect to remote WS server.

During this time, if freeswitch sends any media to proxy, it will get ignored. There is a timer (5s) waiting in initial data payload from freeswitch which if timeout will end the websocket and goes down.

After connected to remote WS, the stream will follows between freeswitch to proxy and gets translated and sent into remote WS.

DTMF events will be relayed by pivot (call's controller) to proxy and will get passed to remote WS if required.

Stream will end when:

1. Freeswitch disconnects (caller hanup or anything happens to FS or connection between proxy and FS)
2. Proxy receives CHANNEL_DESTROY (in case arrives _before_ FS disconnects)
3. Remote WS server disconnects.
4. If unidirectional, the developer stopped the stream (pivot will simply issues the stream stop dialplan to ecallmgr will triggers FS disconnect from proxy.


icehess:
Required PRs:
- master:
    - https://github.com/2600hz/kazoo-pivot/pull/48
    - https://github.com/2600hz/kazoo-core/pull/990
    - https://github.com/2600hz/kazoo-crossbar/pull/633
- 5.4:
    - https://github.com/2600hz/kazoo-pivot/pull/49
    - https://github.com/2600hz/kazoo-core/pull/1035
    - https://github.com/2600hz/kazoo-crossbar/pull/634

icehess (approved):

PR #1005, opened 2024-11-21
KPIV-55: Get media path when using media id for TwiML <Play>

dhuynh-Kazoo:
When passing the media id to the Twiml <Play> verb, it does not work because kazoo could not resolve the file path. So now resolving the media path from media id and current account id


icehess (approved):

dhuynh-Kazoo:
5.4:
[[5.4] KPIV-55: get media path when using media id for TwiML <Play>](https://github.com/2600hz/kazoo-core/pull/1007)

PR #1015, opened 2024-12-02
KZOO-551: parse supplied SIP hosts to extract hostname

jamesaimonetti:
When resource gateways include ports or URI parameters like transport=tls, parse the full value into consituent parts and use just the hostname for SRV and hostname validation.


jamesaimonetti:
Tests in https://github.com/2600hz/kazoo-properly/pull/220

PR #1020, opened 2024-12-05
KZOO-558: improve validation efficiency

kusalfernando:
- remove extra validation check
- report less than 50 attribute conflict errors


kazoo_documents/src/kzd_callflows.erl
```diff
@@ -606,7 +608,18 @@ filter_callflow_list(CallflowId, Callflows) ->
     ].
 
 %%------------------------------------------------------------------------------
-%% @doc Add all `numbers' or `patterns' conflict errors to the
+%% @doc Get sample of conflicting callflows to add conflict errors
+%% @end
+%%------------------------------------------------------------------------------
+-spec conflicting_callflow_sample(kz_json:objects()) -> kz_json:objects().
+conflicting_callflow_sample(CallflowViews) when length(CallflowViews) > 50 ->
```
wdhammika: I noticed that you used a fixed number of 50 in the code for the comparison. Could you clarify the reasoning behind choosing this specific constant value? Is it based on a requirement, or would replacing it with a configurable parameter make sense?

kusalfernando: Nope, it's not a requirement. You are making a good point here, and I will look into that.

kazoo_documents/src/kzd_callflows.erl
```diff
@@ -558,7 +558,9 @@ maybe_validate_attribute_list_is_unique_within_account_callflows(Key, AccountId,
        ) -> kazoo_documents:doc_validation_acc().
 do_validate_attribute_list_is_unique_within_account_callflows(Key, AccountId, CallflowId, {Doc, _Errors}=ValidateAcc) ->
     AccountDb = kzs_util:format_account_db(AccountId),
-    Options = [{'keys', kz_json:get_list_value(Key, Doc, [])}],
+    Attributes = kz_json:get_list_value(Key, Doc, []),
+    Options = [{'keys', Attributes}
+              ,{'limit', kz_datamgr:max_bulk_read()}],
```
danielfinke: Is this going to work if the existing numbers are found across more than `kz_datamgr:max_bulk_read()` different callflows?

kusalfernando: It only checks `max_bulk_read` number of callflows. So should I revert that change? Or do you have any other suggestion?

danielfinke: Yeah I think this could cause problems. Let's say you are trying to update a callflow with numbers 1000-2999 and 4000. The callflow already has the numbers 1000-2999. So the view query will return `max_bulk_read` (2000) results showing the existing callflow, which is later filtered out. But there will be no result included for 4000, even if it exists as another callflow. 4000 will not be rejected for the callflow

Does this seem like a legit issue?

kusalfernando: Yes, it's a legit issue. I will revert that change.

kazoo_documents/src/kzd_callflows.erl
```diff
@@ -606,7 +608,18 @@ filter_callflow_list(CallflowId, Callflows) ->
...
-%% @doc Add all `numbers' or `patterns' conflict errors to the
+%% @doc Get sample of conflicting callflows to add conflict errors
+%% @end
+%%------------------------------------------------------------------------------
+-spec conflicting_callflow_sample(kz_json:objects()) -> kz_json:objects().
+conflicting_callflow_sample(CallflowViews) when length(CallflowViews) > ?CONFLICT_SAMPLE_SIZE ->
+    {FirstFiftyViews, _} = lists:split(?CONFLICT_SAMPLE_SIZE, CallflowViews),
+    FirstFiftyViews;
+conflicting_callflow_sample(CallflowViews) ->
+    CallflowViews.
```
danielfinke: I think this can simply be condensed to
```suggestion
-spec conflicting_callflow_sample(kz_json:objects()) -> kz_json:objects().
conflicting_callflow_sample(CallflowViews) ->
    lists:sublist(CallflowViews, ?CONFLICT_SAMPLE_SIZE).
```

danielfinke (approved):

jamesaimonetti:
@kusalfernando my concern here is there's no proof of this change making a difference. I see no testing available that shows the issue being addressed before this change is made. I would love to see a properly test (or wherever its appropriate) that shows the reason for the ticket in the first place, and when you apply this PR, the test passes as expected.

kusalfernando:
Hi @jamesaimonetti, I was looking to see whether there is any room for improvement. I tried a few options, but I didn't see any improvement, and I got help from Daniel as well. Please let me know if you have any suggestions that I can try out.

PR #1022, opened 2024-12-06
[5.4] KZOO-551: parse supplied SIP hosts to extract hostname

jamesaimonetti:
Backport of #1015

When resource gateways include ports or URI parameters like
`transport=tls`, parse the full value into constituent parts and use
just the hostname for SRV and hostname validation.


jamesaimonetti:
Requires:
- kazoo5: https://github.com/2600hz/kazoo5/pull/182
- properly: https://github.com/2600hz/kazoo-properly/pull/221
- crossbar: https://github.com/2600hz/kazoo-crossbar/pull/628

PR #1029, opened 2024-12-10
KZOO-516: added logic to handle crashing for invalid metaflow schema

sanjeevie6353:
Previously, the system crashed when an incorrect schema body is sent in the API request when trying to activate call recording.  The system sent an a HTTP 500 error. 

This new change will handle that error and will send a HTTP 400 error, stating a client error.

PR #1034, opened 2024-12-13
[5.4] KZOO-516: added logic to handle crashing for invalid metaflow schema

sanjeevie6353:
Previously, the system crashed when an incorrect schema body is sent in the API request when trying to activate call recording. The system sent an a HTTP 500 error.

This new change will handle that error and will send a HTTP 400 error, stating a client error.


wdhammika:
Related master PR - https://github.com/2600hz/kazoo-core/pull/1029

PR #1036, opened 2024-12-13
[5.4] KECA-41: add stream command to control streaming call media to websocket

icehess:
Adding dialplan command for start/stop/pause/resume kz_audio_fork and streaming call's media to a WebSocket.

The stream will the media as raw audio with PCMU (mu-law) encoding. Sample rate and number of channels are configurable.

The stream can be stopped, pause or resume. By default if no Stream-ID is given, these commands will act on all attached audio forks.

* Adding new Stream headers to kapi.dialplan.command.json

* make fmt

* adding blocking functions and rename funs to stream_action

* do not execute now, use correct app name

* please dialyzer

* split audio type to tracks and mix instead of channel-as-stereo

Audust Tracks refers to:
- single: only record the audio that freeswitch is receiving from call's channel (the audio of person that is talking into their phone)
- mixed: record both the audio that freeswitch is receiving also the the audio that freeswitch is also writing into channel (either the audio from other legs or fs is producing/playing on the channel)

Audio Mix indicates how track(s) should be mixed in recroded media:
- mono: all track(s) will be records in single audio channel
- stereo: reading audio will recorded on left channel and writing audio will be recorded on right audio channel.

---------

Co-authored-by: Dat Huynh <dat.huynh@ooma.com>

KECA-41: adjust audio-track header to allow distinct values

Renaming Audio-Track values to make it more distinct which ultimately allows setting it to `outbound` to stream the audio that is being written by FreeSwitch on the channel. Otherwise behaves as before.

- `single` renamed to `inbound`
- `mixed` renamed to `both`
- added `outbound`


icehess:
PRs:
   - master:
       - https://github.com/2600hz/kazoo-core/pull/980
       - https://github.com/2600hz/kazoo-core/pull/1023
       - https://github.com/2600hz/kazoo-ecallmgr/pull/265
       - https://github.com/2600hz/kazoo-ecallmgr/pull/269
   - 5.4:
       - https://github.com/2600hz/kazoo-core/pull/1036
       - https://github.com/2600hz/kazoo-ecallmgr/pull/270

PR #976, opened 2024-10-04
KPIV-51: Add TwiML Dial Conference Record function

dhuynh-Kazoo:
Support Twiml Conference Record:
- Add Recording-Data with URL, method, and recorder
- Add `auto_record_mode` to the conference setup which will set `auto_record` and `min-required-recording-participants` when building a conference profile
- Add `setvar` to kapi_conference for using FS conference `setvar `api

- Some linked PRs:
  https://github.com/2600hz/kazoo-ecallmgr/pull/262
  https://github.com/2600hz/kazoo-conference/pull/55
  https://github.com/2600hz/kazoo-media-mgr/pull/45
  https://github.com/2600hz/kazoo-crossbar/pull/613

Follow up to this PR:
- https://github.com/2600hz/kazoo-core/pull/1033

PR #1037, opened 2024-12-14
[5.4] KPIV-51: Add TwiML Dial Conference Record function (#976)

dhuynh-Kazoo:
Support Twiml Conference Record:
- Add `auto_record` to the conference which will set `auto_record` and `min-required-recording-participants` when building a conference profile
- Add `setvar` to kapi_conference for using FS conference `setvar `api

commits:

* Add TwiML Dial Conference record function
* move recording data to kapps_conference_command, make it binary + check Dial for should record
* cannot set recording data and auto-record, better name for headers
* update schema
* removing auto-record from discovery_req, configure your conference doc instead
* make apis

---------


icehess:
5.4 PRs:
- https://github.com/2600hz/kazoo-core/pull/1037
- https://github.com/2600hz/kazoo-crossbar/pull/635
- https://github.com/2600hz/kazoo-media-mgr/pull/48
- https://github.com/2600hz/kazoo-conference/pull/57
- https://github.com/2600hz/kazoo-ecallmgr/pull/271

PR #1038, opened 2024-12-14
[5.4] KPIV-51: update auto-record conference accessor (#1033)

dhuynh-Kazoo:
Updating the missed change to auto-recrod conference accessor

https://github.com/2600hz/kazoo-core/pull/976


icehess:
added to main pr: https://github.com/2600hz/kazoo-core/pull/1037

PR #1043, opened 2024-12-20
KZOO-116: sometimes allow transactions when billing issues occur

jamesaimonetti:
Many of the client failures are when they think they have money in the bank (and do) and that their bill is OK, but a credit card transaction fails.

We would like it so that the first {DAYS} of the month provide a "grace period", where an account is allowed to go negative to on their monthly service plan amount.

If it is still not resolved by the end of the grace period the account gets cut off.

Also found a failure to save services causing added plans to be lost; updated to ensure the save happens.


jamesaimonetti:
Relevant PRs: 
- https://github.com/2600hz/kazoo-iou/pull/42
- https://github.com/2600hz/kazoo-braintree/pull/21
- https://github.com/2600hz/kazoo-properly-private/pull/23
- https://github.com/2600hz/kazoo-properly/pull/223

PR #1063, opened 2025-01-13
KPIV-47: add ssl option to verify hostname in wildcard certs

icehess:
Adding `customize_hostname_check` option to ssl options for verifying Presented Identifier (e.g hostname) in a peer certificate is in agreement with at least one of the Reference Identifier that the client expects to be connected to.

Using `pkix_verify_hostname_match_fun/1` as matching function to verify wildcard hostnames (`*.cool.com`).


kazoo_stdlib/src/kz_ssl.erl
```diff
@@ -22,6 +22,8 @@ ssl_opts(Options) ->
       [{'versions', ['tlsv1.2']}
       ,{'verify', 'verify_peer'}
       ,{'cacerts', public_key:cacerts_get()} % recommended since OTP-25
+       %% verify matching Presendted Identifier (hostname) with Reference Identifier in wildcard cert
+      ,{'customize_hostname_check', [{'match_fun', public_key:pkix_verify_hostname_match_fun('https')}]}
```
lazedo, line 27: would prefer to have a kz implementation and not a generated function that handles only `https`

icehess: `public_key:pkix_verify_hostname_match_fun/1` doesn't really cares about scheme, it just deals with server reference id(s) and dns name. I think it is just convention to call it https

lazedo: if it doesn't, replace `https` by `no_proto`

icehess: It only accepts `https`. Anyway reading the doc this sufficient enough unless we are dealing with LDAP.

https://github.com/erlang/otp/blob/master/lib/public_key/src/public_key.erl#L1938-L1964

icehess: Also link to doc just in case https://www.erlang.org/doc/apps/public_key/public_key#pkix_verify_hostname_match_fun/1

icehess:
PRs:
- https://github.com/2600hz/kazoo-properly/pull/226
- https://github.com/2600hz/kazoo-tannhauser/pull/18

icehess:
@jamesaimonetti @lazedo 

Any thoughts about updating `kz_http` to insert SSL default options? (James also suggested to refactoring kz_http to use `gun` as `httpc` replacement, though perhaps this could be another PR and discussion)

PR #1066, opened 2025-01-26
KTEL-29: add generic bindings for extending kapi_notifications

dhuynh-Kazoo:
This adds a generic way so out of core applications can extend and add to core KAPI modules.

Apps can call `kapi_definition:bind_api_definition/3` on their startup to bind and call`kapi_definition:unbind_api_definition/4` on their stop to unbind.

These functions can be called the specified Module and Function by KAPI modules to get API definition for the EventName.

The app also must bind to `bind_api_definitions` and unbind variant `unbind_api_definitions` so KAPI module can call them to list all their API definitions provided by that module.

As start, `kapi_notifications` is modified so apps can add extra template api definitions to kapi_notifications without requiring to modify the core. Because of simplicity of `kapi_notifications` it is just enough to only have an API definition for the template id. This API definition can be use for (un/)binding to queue, build, validate and publish notifications. This works with `cb_notifications` too for previewing proposes as it can safely find the req and optional headers and publisher function.

For adding new API definition for a template, apps can call below on app start:

```
start() ->
    ...
    kapi_definition:bind_api_definitions(<<"notifications">>, ?MODULE, 'api_definitions'),
    kapi_definition:bind_api_definition(<<"notifications">>, ?MODULE, 'api_definition'),
    ...

stop() ->
    ...
    kapi_definition:unbind_api_definitions(<<"notifications">>, ?MODULE, 'api_definitions'),
    kapi_definition:unbind_api_definition(<<"notifications">>, ?MODULE, 'api_definition'),
    ...

-spec api_definitions() -> kapi_definition:apis().
api_definitions() ->
    [{MY_APP_TELETYPE_MODULE}:api_definition()].

-spec api_definition(kz_term:ne_binary()) -> kapi_definition:api().
api_definition(?MY_APP_TELETYPE_TEMPLATE_ID) ->
    {MY_APP_TELETYPE_MODULE}:api_definition();
api_definition(_) ->
    'undefined'.


%% later in MY_APP_TELETYPE_MODULE module:
-module(MY_APP_TELETYPE_MODULE).
-spec api_definition() -> kapi_definition:api().
api_definition() ->
    ...
    ...
    ...
    kapi_definition:setters(Setters).
```

Later on the app can use either the generic functions from kapi_definition to build and publish the AMQP message or in case of `kapi_notifications` the app can use shortcut:

```
kz_amqp_worker:cast(Notify, kapi_notifications:notification_publish_fun(<<"template_a">>))
```

The utilities on kapi_definition will try to not crash if the api definition did not found. The publisher fun in kapi_notifications will crash instead if no api definition was found.


dhuynh-Kazoo:
main pr: https://github.com/2600hz/kazoo-hello/pull/48
5.4 backport in: https://github.com/2600hz/kazoo-core/pull/1074

PR #1077, opened 2025-02-10
KTEL-29: reduce log level when notify publusher fun is unknown

icehess:
Since notifpy_publisher notify_sender task module uses only publisher functions from kapi_notification, let's reduce the log level when the publisher func is from other modules.


icehess:
main pr: https://github.com/2600hz/kazoo-hello/pull/48
5.4 backport in: https://github.com/2600hz/kazoo-core/pull/1074

romanat-ooma (approved):

PR #1081, opened 2025-02-14
fixing some invalid JSON schemas

icehess:
it's been long we didn't validate our JSON schemas, some adds invalid addtions. This fixes them.


jamesaimonetti (changes requested):
Make sure the accessor modules are also updated please

PR #1082, opened 2025-02-24
KZOO-578: Moving the recordings to AWS Using Scripting 

manushiperera:
Previously, a bug in the storage plans caused recording attachments to be saved in MODB instead of using the AWS connector settings. While the bug has now been fixed, the recordings already stored in MODB were not automatically moved to AWS. Now, to address this, a new Erlang function has been implemented to locate call recording attachments in all MODBs for a given year and resave them, ensuring they are properly pushed to AWS.


jamesaimonetti:
@manushiperera first thing that jumps out is the mixing of "side-effect" code with "pure" code.

I'd like to see this refactored to put the fetch/save bits in their own functions, and the processing of the data into pure functions.

We can chat more about this is the above isn't clear

lazedo:
`a bug in the storage plans caused recording attachments to be saved in MODB instead of using the AWS connector settings` @manushiperera what `bug` was this ?

manushiperera:
Hi @lazedo , this is regarding `SUPP-582 : Call recordings not being uploaded to AWS successfully` and the commit
https://github.com/2600hz/kazoo-core/commit/99126a62a5270702517e26a2c778ce73cb856bc8

lazedo:
that's what I thought, its not a `bug`, it failed to go into 5.4 because 5.4 was branched out of 5.3 with several code freezes. the original commit is [regex/range should include types for dataplan_type_match_by_classification](https://github.com/2600hz/kazoo-core/commit/6e97ab2197eab4fd34bbf9d50aac939d4093ce76) from Dec 30, 2023

besides that, the specific dataplan was `wrong` for the implementation and could be `fixed` without code change

manushiperera:
Got it @lazedo , that makes sense. Thanks for the explanation.

PR #1086, opened 2025-03-03
KPIV-71: Remove auto-record and recording_data from the kzd_confereces module

dhuynh-Kazoo:
- Conference profile parameter `auto-record` will be fetched from the account db `(configs_conferences doc)` based on the conference profile name, so all functions from kzd_conferences should be removed
- Add Profile-Name to the discovery_req payload


kazoo_amqp/src/api/kapi_conference.erl
```diff
@@ -519,6 +520,7 @@ discovery_req_definition() ->
                ,[{<<"Conference-Doc">>, fun kz_json:is_json_object/1}
                 ,{<<"Conference-ID">>, fun kz_term:is_ne_binary/1}
                 ,{<<"Conference-Name">>, fun kz_term:is_ne_binary/1}
+                ,{<<"Profile-Name">>, fun kz_term:is_ne_binary/1}
```
icehess: The problem with passing these dynamic values are that they are not persisted in the conference doc causing problem down the road. Like imagine node A receives the conference discovery and build/caches the conference profile and asks ecallmgr to start the conference. Then when conf is starting fs asks ecallmgr for profile and ecallmgr sends the config_req to Node B which has no idea of this Profile-Name so it builds a profile using profile on the conference doc or uses the deefault.

Simply put, remove this. Also remove it from kzt_twiml_dial (at least remove profileName). Update twiml documentation instead that if they want auto-record they have to create a new conf profile.

If possible, in kazoo-crossbar or somewhere document some of the profile fields, how to configure (or create profile) in their account (via API) and etc...
Atm no need to document all profile fields. but at least document most relevant one for conference functionality in twiml.

icehess (changes requested):

icehess (approved):

PR #1089, opened 2025-03-04
KZOO-584: allow connections to servers with wildcard TLS certs

jamesaimonetti:
As indicated in the code comment, wildcarded HTTPS certificates are
not accepted with default settings:

```
{error,{tls_alert,{handshake_failure,"TLS client: In state certify at
ssl_handshake.erl:2127 generated CLIENT ALERT: Fatal - Handshake
Failure\n {bad_cert,hostname_check_failed}"}}}
```

Adding this match_fun check allows hostname matching to succeed and
the connection to be established.

OAuth updates to include code_verifier field, if present, when
authenticating the `code` to the OAuth provider.

Use client id from auth app if present, default to name as before


jamesaimonetti:
Schema changes in https://github.com/2600hz/kazoo-crossbar/compare/KZOO-584
Properly changes in https://github.com/2600hz/kazoo-properly/pull/232 (but mostly to support private tests)

lazedo (approved):

PR #1101, opened 2025-03-17
KPIV-37: Update TwiML call status

dhuynh-Kazoo:
Update CallStatus: `ringing, in-progress, busy, failed` when pivot trying to fetching instructions with http request. CallStatus can be included in request queries for GET and request body for POST


romanat-ooma (approved):

icehess (changes requested):
@dhuynh-Kazoo Check all other twiml verbs, make sure if they `kapps_call_command:answer` they also update call status to in-progress.

kazoo_translator/src/converters/kzt_kazoo.erl
```diff
@@ -91,6 +91,7 @@ req_params(Call) ->
       ,{<<"Request-Realm">>, kapps_call:request_realm(Call)}
       ,{<<"To">>, kapps_call:to_user(Call)}
       ,{<<"To-Realm">>, kapps_call:to_realm(Call)}
+      ,{<<"CallStatus">>, kzt_util:get_call_status(Call)}
```
icehess: Remove this. The field in kazoo format is `Call-Status` see line 77.

kazoo_translator/src/kzt_receiver.erl
```diff
@@ -387,7 +387,7 @@ wait_for_offnet(Call, DialProps, DialType) ->
                 ]),
 
     Updates = [fun kzt_util:clear_digits_collected/1
-              ,{fun kzt_util:update_call_status/2, ?STATUS_RINGING}
+              ,{fun kzt_util:update_call_status/2, ?STATUS_IN_PROGRESS}
```
icehess, line 390: Ha good catch, it is indeed must be in-progress. call-status is _incoming_ call status, not the "dialed" call status.

kazoo_translator/src/kzt_util.erl
```diff
@@ -208,7 +208,7 @@ update_call_status(Status, Call) ->
 
 -spec get_call_status(kapps_call:call()) -> kz_term:api_binary().
 get_call_status(Call) ->
-    kapps_call:kvs_fetch(<<"call_status">>, Call).
+    kapps_call:kvs_fetch(<<"call_status">>, ?STATUS_RINGING, Call).
```
icehess: Instead of this, let's add `kzt_util:get_call_status/2` which accepts a default value. Use this in `kzt_kazoo:req_params/1` and `kzt_twiml:req_params/1`, with default for ringing.

kazoo_translator/src/converters/kzt_twiml.erl
```diff
@@ -68,15 +68,18 @@ maybe_set_authorizing_id(Call) ->
...
 exec_elements(Call, [El|Els]) ->
     try exec_element(Call, El) of
         {'ok', Call1} -> exec_elements(Call1, Els);
         {'request', _Call}=REQ -> REQ;
-        {'error', _Call}=ERR -> ERR;
-        {'stop', _Call}=STOP -> STOP
+        {'error', Call1} ->
+            {'error', kzt_util:update_call_status(?STATUS_CANCELED, Call1)};
+        {'stop', Call1} ->
+            {'stop', kzt_util:update_call_status(?STATUS_CANCELED, Call1)}
```
icehess: This will break verbs like hangup, reject, and/or others which set the status to completed. I prefer the verbs and action explicitly set status when stopping.

Check the code and update if necessary .

Reading the code, please also change record verb to return error instead of stop when it encounter error.
Same for Stream, anywhere that it returns error please use error instead of stop.

kazoo_translator/src/converters/kzt_twiml.erl
```diff
@@ -68,15 +68,18 @@ maybe_set_authorizing_id(Call) ->
...
 exec_elements(Call, [#xmlText{}=_El|Els]) ->
     exec_elements(Call, Els);
 exec_elements(Call, [El|Els]) ->
     try exec_element(Call, El) of
         {'ok', Call1} -> exec_elements(Call1, Els);
         {'request', _Call}=REQ -> REQ;
-        {'error', _Call}=ERR -> ERR;
-        {'stop', _Call}=STOP -> STOP
+        {'error', Call1} ->
+            {'error', kzt_util:update_call_status(?STATUS_CANCELED, Call1)};
```
icehess: I prefer to let different verb and action to set this status. Here instead use `kzt_util:update_call_status/2` with default for `failed` in case the action did not set the it when there was an error.

icehess (changes requested):

dhuynh-Kazoo:
master:
[KPIV-37: Update TwiML call status](https://github.com/2600hz/kazoo-core/pull/1101)

[5.4]
[KPIV-37: Update TwiML call status](https://github.com/2600hz/kazoo-core/pull/1124)

PR #1102, opened 2025-03-18
KZOO-589: Update cnam cache mode and allow to flush individual cnam

romanat-ooma:
Cnam being ephemeral value with no way hook to let us know when it expired requires the default cache type to be peek. In same cases cnam never expires and cache becomes stale, we want to have a predictable expiry on cnam fetch. Also able to remove individual cnam from cache when needed.


lazedo:
@romanat-ooma `we want to have a predictable expiry on cnam fetch` why is this ? rolling expires are what we need, lets not make changes that only benefit tests and monitoring

romanat-ooma:
@lazedo reverted to fetch being default, the issue arise for cache time any longer than few hours. A single call from customer after few hours will bump the cache keeping the cached value indefinitely. Will add documentation to explain when to use peek.

romanat-ooma:
CI depends on https://github.com/2600hz/kazoo-crossbar/pull/673

icehess (approved):

PR #1104, opened 2025-03-19
[5.3] KZOO-590 When an Internal extension calls a User/Device that forwards to and external number with call is sent out with ext number instead of a valid CID

eezhov-ooma:
Port fix for call-forward CID from https://github.com/2600hz/kazoo-core/commit/9563bda18420c366e8a74e7bcded6edd1b9e0daa (master, 5.4)


lazedo:
@eezhov-ooma why is [this](https://github.com/2600hz/kazoo-core/commit/b82a6359df698ee403fbab951befac27573d5e8b) being reverted ? this will bring back [fixed issue](https://oomacorp.atlassian.net/browse/KZOO-481]) here.

eezhov-ooma:
@lazedo without this changes we see the issue described in SUPP-662 "When an internal extension calls a user that has a forward set on them or has a Device that rings to an external number, the call fails when  "Keep Original Caller-ID" is enabled. Logs show the call going out with the internal extension number of the calling party being sent instead of of the Account default CID. Unable to replicate on Sandbox and Zswitch"

lazedo:
@eezhov-ooma I don't think so. maybe customer is on a different release ? make sure you don't invalidate KZOO-481 fix which replaced what you're trying to put back

lazedo:
fixed by https://github.com/2600hz/kazoo-core/pull/1115 and https://github.com/2600hz/kazoo-core/pull/1116

PR #1105, opened 2025-03-19
KWEB-15: Handling empty objects and list when encoding JSON to QS

artable:
Linked PRs:
https://github.com/2600hz/kazoo-webhooks/pull/50
https://github.com/2600hz/kazoo-properly/pull/233

Previously kz_http_util would include k/vs with empty json and empty list values  when encoding, which would add extra "&" for each key value. Updating `kz_http_util:json_to_querystring\1` to solve this.


icehess:
@artable You need rebase to latest master, address the dialyzer issue and fmt (trailing whitespace)

icehess (approved):

PR #1123, opened 2025-04-01
KWEB-14: (Reliably) send account creation and deletion webhooks

artable:
Linked PRs:
https://github.com/2600hz/kazoo-crossbar/pull/680 (see this PR for description)
https://github.com/2600hz/kazoo-webhooks/pull/56
https://github.com/2600hz/kazoo-teletype/pull/106

adds an `account_deleted` notification.


kazoo_amqp/src/api/kapi_notifications.erl
```diff
@@ -468,6 +473,39 @@ new_account_definition() ->
               ,{fun kapi_definition:set_binding/2, ?BINDING_STRING(Category, <<"new">>)}
               ,{fun kapi_definition:set_restrict_to/2, 'new_account'}
               ,{fun kapi_definition:set_required_headers/2, [<<"Account-ID">>]}
+              ,{fun kapi_definition:set_optional_headers/2, [<<"Account-API-Key">>
+                                                            ,<<"Account-Name">>
+                                                            ,<<"Account-Realm">>
+                                                            ,<<"No-Email">>
```
jamesaimonetti: No-Email leaves a lot to be desired in the naming department.

`Should-Notify` suggests a `boolean()` value, and historically the `notify` app was the predecessor to `teletype`, and teletype can send notifications via HTTP, not just email

jamesaimonetti (changes requested):

kazoo_amqp/src/api/kapi_notifications.erl
artable: I wasn't particularly happy with `<<"No-Email">>`. The one thing I like about it is it suggests of itself the default, teletype (and whatever else) will send a notification. `<<"Should-Not-Override">>` is just awkward though. Ah well, I'll change it to <<"Should-Notify">>

kazoo_amqp/priv/couchdb/schemas/kapi.notifications.new_account.json
```diff
@@ -45,12 +45,18 @@
         "HTML": {
             "type": "string"
         },
+        "No-Email": {
```
icehess: What is this new property? where is gets used?

kazoo_amqp/priv/couchdb/schemas/kapi.notifications.new_account.json
```diff
@@ -45,12 +45,18 @@
...
+            "type": "string"
+        },
         "Preview": {
             "type": "boolean"
         },
         "Reply-To": {
             "type": "string"
         },
+        "Should-Notify": {
```
icehess, line 54: Can you explain why this change is needed?

kazoo_amqp/src/api/kapi_notifications.erl
```diff
@@ -49,6 +49,11 @@
         ,publish_new_account/1
         ,publish_new_account/2
         ]).
+-export([account_deleted/1
```
icehess, line 52: I'm curious to know what other approaches are tried and failed for account deletion? Why using notification was chosen?

icehess (changes requested):
Explain why notification for account deletion is used instead of like listening for changes on `accounts` database?

Also the Pr title says `creation` too, but i don't see anything in that regard.

kazoo_amqp/priv/couchdb/schemas/kapi.notifications.new_account.json
artable: Oh this should be removed, James suggested changing the name to "Should-Notify", which I also prefer.

kazoo_amqp/priv/couchdb/schemas/kapi.notifications.new_account.json
artable: See other comments! :)
In short, so that crossbar always sends `account_created` notifications (so webhooks can always get them), and `teletype` can choose not to send the notif.

kazoo_amqp/src/api/kapi_notifications.erl
artable: See Crossbar PR. Soft deleting the doc does 'work` for objects, though It has some side effects and I am concerned there are more that I didn't see.

kazoo_amqp/src/api/kapi_notifications.erl
```diff
@@ -468,6 +473,39 @@ new_account_definition() ->
...
               ,{fun kapi_definition:set_restrict_to/2, 'new_account'}
               ,{fun kapi_definition:set_required_headers/2, [<<"Account-ID">>]}
+              ,{fun kapi_definition:set_optional_headers/2, [<<"Account-API-Key">>
+                                                            ,<<"Account-Name">>
+                                                            ,<<"Account-Realm">>
+                                                            ,<<"Should-Notify">>
+                                                            | ?DEFAULT_OPTIONAL_HEADERS
+                                                            ]}
+              ,{fun kapi_definition:set_values/2, ?NOTIFY_VALUES(EventName)}
+              ,{fun kapi_definition:set_types/2, []}
```
icehess: Two suggestion, rename `SHOULD_NOTIFY` to `SHOULD_SEND_EMAIL` to represent it better (if  get should-notify in webhook its look weird, right?). Also if you add types for this key, then it validates during runtime that it is indeed in the correct type, also we create a proper JSON schema with correct type for this header:

```suggestion
                                                            ,<<"Should-Send-Email">>
                                                            | ?DEFAULT_OPTIONAL_HEADERS
                                                            ]}
              ,{fun kapi_definition:set_values/2, ?NOTIFY_VALUES(EventName)}
              ,{fun kapi_definition:set_types/2, [
               ,[{<<"Should-Send-Email">>, fun kz_term:is_boolean/1}
                ]
               }
```

icehess (changes requested):

icehess (approved):

icehess (approved):

icehess:
@jamesaimonetti If you are happy with changes, i think you need to approve as github is avoiding me to merge this.

jamesaimonetti (approved):

PR #1130, opened 2025-04-07
KZOO-591: Disable telemetry by default

romanat-ooma:
Disable kazoo-telemetry as we have shut down waveguide servers.


dhuynh-Kazoo (approved):

romanat-ooma:
rewritten to make sure we disable only if servers are not running.

PR #1133, opened 2025-04-08
KBLA-22 High Load on blackhole server

eezhov-ooma:
Linked review:
- https://github.com/2600hz/kazoo-blackhole/pull/63
- https://github.com/2600hz/kazoo-core/pull/1133

---

Customer is still seeing high load on bh server. 
From the logs there are lot of crashes on `gen_server:call`, .e.g.

All crashes are related to the `kz_hooks_listener` across provided logs. I suggest porting "BIZ-7701: blackhole, remove unnecessary gen_listener calls to improve the blackhole performance" to 2600hz code base. This fix reduce number of gen_server:call calls under the `kz_hooks_listener`. 

See details in the ticket.

Previous changes were under https://github.com/2600hz/kazoo-blackhole/pull/52


romanat-ooma (approved):

kazoo_hooks/src/kz_hooks_listener.erl
```diff
@@ -217,3 +219,25 @@ binding_to_call_event({<<"call">>, CallProps}, CallEvents) ->
         Events -> lists:usort(Events ++ CallEvents)
     end;
 binding_to_call_event({_Binding, _BindProps}, CallEvents) -> CallEvents.
+
+-spec is_new_event('all' | kz_term:ne_binary()) -> boolean().
+is_new_event('all') -> 'true';
+is_new_event(EventName) ->
+    case erlang:get('process_name') == 'blackhole_listener'
```
jamesaimonetti: this is not acceptable; the core libs have no idea what a `blackhole_listener` is (and should not know). This needs to be fixed in a different way

jamesaimonetti (changes requested):

icehess (changes requested):
I kind of don't like handling specific gen_lisnter this way. There should be a common way so other listeners can use to avoid problem like this.

The underlying problem here that hook register is also using `call` instead of `cast`. Have you consider to add options to kazoo hook how it register so it uses cast instead of call?

kazoo_hooks/src/kz_hooks_listener.erl
jamesaimonetti: using process dictionary for this is also a non-starter. there has to be a really really good reason for using it and dynamic data like this is not appropriate

eezhov-ooma: @jamesaimonetti I can remove `blackhole_listener` form this function, then we will check it for all apps
```
case erlang:get('hook_events') of
        'undefined' -> erlang:put('hook_events', []), 'true'; 
        Events -> not lists:member(EventName, Events)
    end.
```
anyway it will be a quite small list of call events, like "CHANNEL_CREATE".

PR #1135, opened 2025-04-09
KZOO-576: Add deprecated read and write values to kazoo 

romanat-ooma:
During bigcouch upgrade from 0.4.x we lost the ability to set read and write values for a document fetch and post from database. This commit allows us get that ability back.


kazoo_couch/src/kz_couch_doc.erl
```diff
@@ -51,20 +51,32 @@ open_doc(Conn, DbName, DocId) ->
           doc_return().
 open_doc(#server{}=Conn, DbName, DocId, Options) ->
     Db = get_db(Conn, DbName),
-    do_fetch_doc(Db, DocId, Options).
+    do_fetch_doc(Db, DocId, maybe_read_option(Options)).
+
+maybe_read_option(Options) ->
+    case kz_config:get_integer(kz_dataconfig:get_config_section(), <<"read">>) of
```
icehess: Maybe first check if `Option` has `r` (or `w`) set and use that, then default to get it from config?

Also probably allow Options to turn off setting r and w. Like if it is set to `null` don't do anything to give callers intentionally turn off insertion of these parameters if they need it.

icehess: And also obviously use the get_read and get_write from maintenance module :)

kazoo_couch/src/kazoo_couch_maintenance.erl
```diff
@@ -12,4 +12,55 @@
...
+
+-define(SECTION_CONFIG, kz_dataconfig:get_config_section()).
+
+-spec get_read() -> 'ok'.
+get_read() -> get_config_val(<<"read">>).
+
+-spec get_write() -> 'ok'.
+get_write() -> get_config_val(<<"write">>).
+
+get_config_val(Config) ->
```
icehess: better name for variable:

```suggestion
get_config_val(Key) ->
```

kazoo_couch/src/kazoo_couch_maintenance.erl
```diff
@@ -12,4 +12,55 @@
...
+    Section = ?SECTION_CONFIG,
+    case kz_config:get_integer(Section, Config) of
+        [] ->
+            io:format("~s value empty in section ~s ~n", [Config, Section]);
```
icehess: maybe return a default value?

kazoo_couch/src/kazoo_couch_maintenance.erl
```diff
@@ -12,4 +12,55 @@
...
+        [Value] ->
+            io:format("~s value: ~p ~n", [Config, Value])
```
icehess: shouldn't it return the Value?

kazoo_couch/src/kazoo_couch_maintenance.erl
icehess: Also no need for log line perhaps?1

kazoo_couch/src/kz_couch_doc.erl
```diff
@@ -51,20 +51,32 @@ open_doc(Conn, DbName, DocId) ->
...
 -spec save_docs(server(), kz_term:ne_binary(), kz_json:objects(), kz_term:proplist()) ->
           {'ok', kz_json:objects()} |
           couchbeam_error().
 save_docs(#server{}=Conn, DbName, Docs, Options) ->
     Db = get_db(Conn, DbName),
-    do_save_docs(Db, Docs, Options).
+    do_save_docs(Db, Docs, maybe_write_option(Options)).
+
+maybe_write_option(Options) ->
+    case kz_config:get_integer(kz_dataconfig:get_config_section(), <<"write">>) of
```
icehess: Same comment as for read, allow caller to set this value or request to not this value get set.

icehess (changes requested):

kazoo_couch/src/kz_couch_doc.erl
romanat-ooma: @icehess db options does have read and write values but hot loading the values was causing issues, so that's why kz_config is the only place we read values from.

get_read, get_write are there in maintenance, for unsetting value, unset_read and unset_write is present that's why not setting it null here. Removing the values from config has same effect.

kazoo_couch/src/kazoo_couch_maintenance.erl
romanat-ooma: @icehess default value is not returned to keep it unset when no value is present.

log line is present to inform sup command user. will add docs to go over the usage.

kazoo_couch/src/kazoo_couch_maintenance.erl
romanat-ooma: @icehess just a maintenance command so return ok.

kazoo_couch/src/kz_couch_doc.erl
icehess: Sorry I didn't meant to read the config from system_config :)) that obviously won't work.

The `Options` here are the view options passed by caller and kz_datamgr, like `kz_datamgr:open_doc(Db, DocId, [{r, null}]).`

I'm just saying that allow the caller to set `r` and `w`, and if it is `null`, do not add any r and w. if undefined then set the value them from config.

Regarding `kazoo_couch_maintenance:get_read/0` and get_write. Either move them to `kz_couch_config` or let them return the value as it. 

There is no need to repeat `kz_config:get_integer(kz_dataconfig:get_config_section(), Key)` twice both in kz_couchdb_* and maintenance.


Also Does view accepts `r` too? you can added it there too

kazoo_couch/src/kz_couch_doc.erl
romanat-ooma: Will explain maintenance command usage to make this possible.

kazoo_couch/src/kz_couch_doc.erl
romanat-ooma: @icehess `kazoo_couch_maintenance:get_read/0` are for maintenance commands only to check what is the currently value, the values are read directly from kapps config. I tried using config coming from couch connection in options but it doesn't change when we try to change kapps config value without restarting.

lazedo:
@romanat-ooma please open a separate pr for view/ken stuff

lazedo (changes requested):
split the pr fr the ken/view stuff

kazoo_couch/src/kz_couch_doc.erl
lazedo: iirc, each time you read something with kz_config its going to parse ini file.

lazedo: and I don't see a value in this pr but to `trick` the system. allowing per request could be ok, but for all requests seems you're forcing a behaviour that may be masquerading a problem that will never get fixed

romanat-ooma: @lazedo I can read the same values from `#server{}=Conn`, would that be better The only thing I am not able to do with it is inject values at runtime, kz_config was to accommodate change of values on runtime without restarting kazoo ? I agree this is couchdb problem but my testing has revealed a lot of gains if we set r=1 so I believe this is a valid pr.  Also this is totally optional setting which behaves as default when not present.

romanat-ooma:
> @romanat-ooma please open a separate pr for view/ken stuff

created separate ticket for it. https://github.com/2600hz/kazoo-core/pull/1136

kazoo_couch/src/kz_couch_doc.erl
lazedo: with r=1 you run the risk of getting the wrong version unless its a standalone couchdb

romanat-ooma: @lazedo view lookup should also give us wrong data as r=1 for all view lookups. so I believe it is a rare case where it would happen. Keeping requests to local zone couches can also make sure the requests get the latest data. while keeping latency low.  in my tests I haven't gotten bad data although I haven't tested with very busy systems.

kazoo_couch/src/kz_couch_doc.erl
```diff
@@ -51,20 +51,32 @@ open_doc(Conn, DbName, DocId) ->
...
 open_doc(#server{}=Conn, DbName, DocId, Options) ->
     Db = get_db(Conn, DbName),
-    do_fetch_doc(Db, DocId, Options).
+    do_fetch_doc(Db, DocId, maybe_read_option(Conn, Options)).
+
+maybe_read_option(#server{options=ConnOpts}, Options) ->
+    case proplists:get_value('read', ConnOpts) of
+        'undefined' -> Options;
+        Value -> [{'r', Value} | Options]
+    end.
```
icehess:
```suggestion
    case props:get_value('read', Options, props:get_value('read', ConnOpts) of
        'null' -> Options; %% allow kz_datamgr:open_doc/3 users to disable 'read'
        'undefined' -> Options;
        Value -> [{'r', Value} | props:delete('read', Options)]
    end.
```

kazoo_couch/src/kz_couch_doc.erl
```diff
@@ -51,20 +51,32 @@ open_doc(Conn, DbName, DocId) ->
...
 save_docs(#server{}=Conn, DbName, Docs, Options) ->
     Db = get_db(Conn, DbName),
-    do_save_docs(Db, Docs, Options).
+    do_save_docs(Db, Docs, maybe_write_option(Conn, Options)).
+
+maybe_write_option(#server{options = ConnOpts}, Options) ->
+    case proplists:get_value('read', ConnOpts) of
+        'undefined' -> Options;
+        Value -> [{'w', Value} | Options]
+    end.
```
icehess: Isthe key name `write` or `read`?

anyway:
```suggestion
maybe_write_option(#server{options = ConnOpts}, Options) ->
    case props:get_value('write', Options, props:get_value('write', ConnOpts)) of
            'null' -> Options; %% allow kz_datamgr:save_doc/3 users to disable 'read'
        'undefined' -> Options;
        Value -> [{'w', Value} | props:delete('write', Options)]
    end.
```

icehess (changes requested):
Can also write a doc how to configure this and a sample? a brief explaining of options and their consequences/results would be nice too.

romanat-ooma:
Working on resolving some issues with move to using data plans.

romanat-ooma:
closing this and continuing on KZOO-613

PR #1136, opened 2025-04-11
KZOO-576: Disable auto indexing of couchdb views

romanat-ooma:

Depends on: https://github.com/2600hz/kazoo-crossbar/pull/688

This pr adds `autoupdate: false` to couch views. As we are used to disabling ken for background indexing https://docs.couchdb.org/en/stable/config/indexbuilds.html

Docs:
1: autoupdate: false
This is to disable background indexing of couch views.

System config db will now contain documents to configure this value.

`design_docs` doc will contain the json

```json
{
  "id": "design_docs",
  "default": {
    "options": {
       "autoupdate": false
      },
      "classifications": {
        "system": {
          "options": {
            "autoupdate": true
          }
        }
      },
      "views": {
        "system": {
          "system_configs": {
            "options": {
              "autoupdate": false
            }
          }
        }
      }
    }
}
```

The json is nested with following values.

```json
{
  "id": "design_docs",
  "default": {
    "classifications": {
      "{CLASSIFICATION}": {
        "options": {}
      }
    },
    "views": {
      "{classification}": {
        "{design_name}": {
          "options": {}
        }
      }
    }
  }
}
```

Here design name is without the `_design/` part e.g. `_design/accounts` will be `accounts`

to refresh views `sup kapps_maintenance refresh` can be used.


kazoo_data/src/kz_datamgr.erl
```diff
@@ -1882,5 +1882,32 @@ view_definitions(<<DbName/binary>>, <<Classification/binary>>) ->
         {'error', _} -> [];
         {'ok', JObjs} ->
             ViewDefs = [kz_json:get_json_value(<<"view_definition">>, JObj) || JObj <- JObjs],
-            [{kz_doc:id(ViewDef), ViewDef} || ViewDef <- ViewDefs]
+            [{kz_doc:id(ViewDef), maybe_add_design_options(ViewDef, Classification)} || ViewDef <- ViewDefs]
     end.
+
+-spec maybe_add_design_options(kz_json:object(), kz_term:ne_binary()) -> kz_json:object().
+maybe_add_design_options(View, Classification) ->
+    Id = kz_doc:id(View),
```
icehess: You you are doubling `design_doc._design/accounts`?

kazoo_data/src/kz_datamgr.erl
```diff
@@ -1882,5 +1882,32 @@ view_definitions(<<DbName/binary>>, <<Classification/binary>>) ->
...
+    case kapps_config:get_json(<<"design_doc.", Id/binary>>, <<"design_options">>, 'undefined') of
```
icehess: tbh having dealing with `/` in doc id is painful, i rather have `kapps_config:get_json(<<"design_docs">>, [<<"designs">>, kz_doc:id(View)], [undefined]`, any thoughts? 

Also more importantly  here actually need more info, because design doc id can be in multiple dbs with same id. look at register_view for example.

icehess (changes requested):

kazoo_data/src/kz_datamgr.erl
romanat-ooma: @icehess yes, if only using design_doc.accounts, it clashes with classifications, so to keep them different I had to use full id.

kazoo_data/src/kz_datamgr.erl
romanat-ooma: @icehess the intention is to just base it in design doc id, which will be present in multiple databases. Not sure if we can do anything else with clashing names. classification and doc id has the same issue.

kazoo_data/src/kz_datamgr.erl
```diff
@@ -1882,5 +1882,32 @@ view_definitions(<<DbName/binary>>, <<Classification/binary>>) ->
...
+        'undefined' ->
+            maybe_add_classification_options(View, Classification);
+        ViewOptions ->
+            kz_json:merge(View, ViewOptions)
+    end.
+
+-spec maybe_add_classification_options(kz_json:object(), kz_term:ne_binary()) -> kz_json:object().
+maybe_add_classification_options(View, Classification) ->
+    case kapps_config:get_json(<<"design_doc.", Classification/binary>>, <<"design_options">>, 'undefined') of
```
jamesaimonetti: I was thinking more like one system_config doc like:

```json
{"id":"design_docs"
 ,"default":{
     "design_options":{
        "default":{...}
        ,"{CLASSIFICATION}":{...}
        ,"{DESIGN_DOC_ID}":{...}
    }
 }
}
```

icehess: second that :)

icehess: Or perhaps:

```
{
  "id": "design_docs",
  "default": {
    "options": {},
    "classifications": {
      "{CLASSIFICATION}": { "options": {}}}
    },
    "views": {
      "{classfiication}": {
        "{design_name}": { "options": {}}}
      }
    }
  }
}
```

icehess: uodated fyi ^

jamesaimonetti: wfm (plz document in kazoo_data/doc somewhere Taran)

romanat-ooma: on it!! Thank you.

kazoo_data/src/kz_datamgr.erl
icehess: see James' comment about single doc

kazoo_data/src/kz_datamgr.erl
romanat-ooma: Code changed as requested

romanat-ooma: added documentation

icehess:
PRs:
- master:
    - https://github.com/2600hz/kazoo-crossbar/pull/688
    - https://github.com/2600hz/kazoo-core/pull/1136
- 5.4:
    - wip

icehess (approved):

PR #1137, opened 2025-04-12
KPIV-52: New pivot utility document funs for pivot schema

dhuynh-Kazoo:
* `/pivot` with GET/PUT
* `/pivot/pivotId` with GET, PUT, PATCH, DELETE
- add `pivot` to `by_type_id` for crossbar GET `/pivot`
- new `/_design/pivot/_view/crossbar_listing` to fetch all pivot document in db


kazoo_documents/src/kzd_pivot.erl
```diff
@@ -0,0 +1,207 @@
...
+-export([name/1, name/2, set_name/2]).
+-export([req_body_format/1, req_body_format/2, set_req_body_format/2]).
+-export([req_format/1, req_format/2, set_req_format/2]).
+-export([req_timeout_ms/1, req_timeout_ms/2, set_req_timeout_ms/2]).
+-export([serverless/1, serverless/2, set_serverless/2]).
+-export([serverless_flow_doc/1, serverless_flow_doc/2, set_serverless_flow_doc/2]).
+-export([serverless_flow_type/1, serverless_flow_type/2, set_serverless_flow_type/2]).
+-export([skip_module/1, skip_module/2, set_skip_module/2]).
+-export([voice_url/1, voice_url/2, set_voice_url/2]).
+-export([type/0]).
```
icehess: tbh i prefer `pvt_type`:

```suggestion
-export([pvt_type/0]).
```

kazoo_apps/priv/couchdb/account/pivot.json
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "language": "javascript",
+    "views": {
+        "crossbar_listing": {
+            "map": [
+                "function(doc) {",
+                "  if (doc.pvt_type != 'pivot' || doc.pvt_deleted) return;",
+                "  emit(doc._id, {",
+                "    'name': doc.name,",
+                "    'number': doc.number,",
```
icehess: Why number? Pivot doc don't have a number. number must be defined on callflow as before.

kazoo_apps/priv/couchdb/account/pivot.json
```diff
@@ -0,0 +1,30 @@
...
+                "    'method': doc.method,",
+                "    'req_timeout': doc.req_timeout,",
```
icehess: Better to not add too much stuff, i think just name, url (rename voice_url to url maybe??), and method are enough.

This will not increase index file size.

icehess (changes requested):

kazoo_apps/priv/couchdb/account/pivot.json
icehess: Remove `number` please. Pivot doc does not have number.

kazoo_apps/priv/couchdb/account/pivot.json
icehess: Remove unnecessary fields from view. I think for now we only need:
- name
- url
- method

icehess (changes requested):

PR #1138, opened 2025-04-15
KBLA-25: federate kz_events for call events

jamesaimonetti:
Currently only two apps make use of kz_hooks for call events: blackhole and jonny5.

Both apps require federated call events.

Currently no users of kz_events:bind_call_id/{1,2} use a wildcarded call-id (which blackhole/jonny5 would use) so no changes to their operation.

This change should not increase current inter-zone channel events as kz_hooks already causes those events to be delivered; this just changes the mechanism to the newer shared kz_events - one copy of the event from AMQP to the VM; 0 or more copies shared among bound processes on the local VM.


lazedo (approved):

kazoo_couch/src/kz_couch_util.erl
```diff
@@ -236,7 +236,7 @@ server_info(#server{}=Conn) -> couchbeam:server_info(Conn).
 -spec membership(server()) -> kz_json:object().
 membership(#server{url=ServerUrl}) ->
     URL = hackney_url:make_url(ServerUrl, <<"_membership">>, []),
-    case kz_http:get(URL) of
+    case kz_http:get(URL, [], [{'log', 'false'}]) of
```
icehess, line 239: thank you so much for this!!

PR #1141, opened 2025-04-16
[5.4] KZOO-576: Disable auto indexing of couchdb views

romanat-ooma:

This pr adds `autoupdate: false` to couch views. As we are used to disabling ken for background indexing https://docs.couchdb.org/en/stable/config/indexbuilds.html

Docs:
1: autoupdate: false
This is to disable background indexing of couch views.

System config db will now contain documents to configure this value.

`design_docs` doc will contain the json

```json
{
  "id": "design_docs",
  "default": {
    "options": {
       "autoupdate": false
      },
      "classifications": {
        "system": {
          "options": {
            "autoupdate": true
          }
        }
      },
      "views": {
        "system": {
          "system_configs": {
            "options": {
              "autoupdate": false
            }
          }
        }
      }
    }
}
```

The json is nested with following values.

```json
{
  "id": "design_docs",
  "default": {
    "classifications": {
      "{CLASSIFICATION}": {
        "options": {}
      }
    },
    "views": {
      "{classification}": {
        "{design_name}": {
          "options": {}
        }
      }
    }
  }
}
```

Here design name is without the `_design/` part e.g. `_design/accounts` will be `accounts`

to refresh views `sup kapps_maintenance refresh` can be used.


romanat-ooma:
PRs:

master:
https://github.com/2600hz/kazoo-crossbar/pull/688
https://github.com/2600hz/kazoo-core/pull/1136

5.4:
https://github.com/2600hz/kazoo-crossbar/pull/690
https://github.com/2600hz/kazoo-core/pull/1142

PR #1147, opened 2025-04-17
KZOO-612: Prepare for OTP27

romanat-ooma:
In preparation for OTP27 handling of deprecated functions.

`code:lib_dir/2` was deprecated and using `filename:join(code:lib_dir/1))` is right approach.


kazoo_fixturedb/src/kz_fixturedb_server.erl
```diff
@@ -120,7 +120,7 @@ maybe_use_app_connection(#{options := Options}=Server, DbName) ->
 set_app_connection(#{options := Options}=Server, AppName) ->
     Path = case maps:get('test_db_subdir', Options, 'undefined') of
                'undefined' -> code:priv_dir(AppName);
-               P -> code:lib_dir(AppName, kz_term:to_atom(P, 'true'))
+               P -> filename:join(code:lib_dir(AppName), kz_term:to_atom(P, 'true'))
```
lazedo: this line looks incongruent with the rest of the changes as it maintains `atom` as second arg

lazedo (changes requested):

kazoo_fixturedb/src/kz_fixturedb_server.erl
romanat-ooma: thanks @lazedo for the review, fixing it.

lazedo (approved):

PR #1150, opened 2025-04-23
KZOO-615 CNAM gets enabled when touching (saving without making chang…

enoskov-at-ooma:
Add a sup command which queries all number docs and disables cnam features on all Cluster wide.


kazoo_numbers/src/kazoo_numbers_maintenance.erl
```diff
@@ -2112,3 +2114,54 @@ migrate_port_requests() ->
...
+            _ = disable_cnam([kz_doc:id(JObj) || JObj <- JObjs]),
+            timer:sleep(?TIME_BETWEEN_ACCOUNTS_MS),
+            disable_cnam_features(NumberDb, Offset + Length);
+        {'error', _R} ->
+            ?SUP_LOG_DEBUG("failed to get DIDs from ~s: ~p", [NumberDb, _R])
+    end.
+
+-spec disable_cnam(kz_term:ne_binaries()) -> knm_numbers:return().
+disable_cnam(DIDs) ->
+    Update = kz_json:set_values([{[<<"cnam">>,<<"display_name">>], <<>>},
```
jamesaimonetti: should this use `null` instead of empty binary? Do we want to remove the `cnam.display_name` key/value pair entirely? or set it to empty?

enoskov-at-ooma: @jamesaimonetti [<<"cnam">>,<<"display_name">>], null or undefined doesn't work for update.
Empty string also removes the feature and MonsterUI shows it as disabled.

I can offer the following code
<<"cnam">>, kz_json:new() which makes a number doc looking as follow:

 "cnam": {} 
 
 and also removes features. I just don't want to change knm_phone_number:update_doc() for this relatively small task and empty string works.

enoskov-at-ooma: btw, via API it is also not possible to set it to null
`curl -X PATCH -H "X-Auth-Token: $AUTH_TOKEN" -H "Content-Type: application/json" "https://crocubot.2600hz.dev:8443/v2/accounts/c8ec5aab18c4a2e32d7a5b8730b9ea7f/phone_numbers/%2B15852058907" -d '{"data":{"cnam":{"display_name":null}}}' | python3 -mjson.tool`

"message": "Value did not match type(s): string"

romanat-ooma: @enoskov-at-ooma I tested the code and it disabled the cnam but any operation afterwards would cause error due to empty display_name, I would prefer the display name to be removed completely. To test out the change after cnam is disabled on a number, try to change its carrier module or even trying saving with same carrier, if it doesn't error out that means the operation is safe.

romanat-ooma:
After confirming with OPS and due to not being able to properly disable outbound cnam I have changed the code to just disable inbound cnam and changed function names to reflect the same. Will wait for CI to pass before merging.

romanat-ooma (approved):
tested and its disabling cnam on cluster

PR #1170, opened 2025-05-08
KZOO-564: add 'and'/'not' types to match lists

jamesaimonetti:
Process temporal rule sets as well


jamesaimonetti:
Schema updates in https://github.com/2600hz/kazoo-crossbar/pull/701

jamesaimonetti:
Requires schema changes: https://github.com/2600hz/kazoo-crossbar/pull/701
Some testing: https://github.com/2600hz/kazoo-properly/pull/246

PR #1182, opened 2025-05-21
KZOO-626: check /tmp and reupload recording if found

jamesaimonetti:


romanat-ooma (approved):

kazoo_attachments/src/kazoo_attachments_maintenance.erl
```diff
@@ -0,0 +1,132 @@
...
+-module(kazoo_attachments_maintenance).
+
+-export([retry_failed_uploads/0
+        ,retry_failed_account/1
+        ,retry_failed_account_modb/3
+        ]).
+
+-elvis([{elvis_style, no_debug_call, disable}]).
+
+-define(SUMMARY_VIEW, <<"recordings/crossbar_listing">>).
```
icehess, line 20: What about other attachment types? voicemail? fax? Is recording the only one support for now?

icehess (approved):

PR #1190, opened 2025-05-26
KZOO-640: Create kazoo replicator

romanat-ooma:
Create kazoo replicator which uses couchdb replication to perform replications between couchdb clusters.

# KAZOO REPLICATOR

Replicate the data between two couchdb nodes.

## HOW IT WORKS

we can pass some arguments and let the replication begin

start by creating a replication job  
`sup replicator create`  
`sup replicator create optionalname`

it will return a job id, use that job id to fill out other parameters for db

### Fill out replication job info

```bash
sup replicator set $id from_host $COUCHURL
sup replicator set $id from_username $USERNAME
sup replicator set $id from_password $PASSWORD
sup replicator set $id from_port default to 5984
sup replicator set $id to_host $COUCHURL
sup replicator set $id to_username $USERNAME
sup replicator set $id to_password $PASSWORD
sup replicator set $id to_port default to 5984
sup replicator set $id worker_node to
sup replicator set $id continuous true
sup replicator add $id $masteraccountid
sup replicator add $id $reselleraccountid
sup replicator add $id $accountid3
```

* From CouchURL - url or ip of couch node. db1.kazoo.aio  
* From username - username of from node  
* From password - password of from node  
* To CouchURL - url or ip of couch node. 10.62.11.100  
* To and From port - if not using Haproxy no need to add value for ports they will be autopopulated.  
* Worker Node - Node which will push or pull the replication jobs. Usually To node which is node we are copying data to.  
The value will be from when we want to create job which does master master replication back to our main db cluster from the backup server.  
It is possible to have a third worked node carry on all the replication jobs but that isn't currently supported.  
* Continuous - defaults to false, wether to keep sync alive or oneshot sync.  

#### Account based replication
In case where only replication of few accounts is needed

Make sure any master or reseller account leading up to replicated accounts is used.
By default all system dbs will be replicated.

```bash
sup replicator add $id $masteraccountid
sup replicator add $id $reselleraccountid
sup replicator add $id $accountid3
```

### Other commands to interact with jobs

`sup replicator status`
show the current status for all the replication jobs

`sup replicator status id`
show the current status for all replications for a particular job

`sup replicator stop id`
stop the current replication

`sup replicator start id`
start the selected replication

`sup replicator update`
update job with latest dbs

`sup replicator delete id`
delete the complete replication job

`sup replicator stop_all`
stop all replication jobs

`sup replicator delete_all`
delete all replication jobs


kazoo_replicator/src/replicator.erl
```diff
@@ -0,0 +1,183 @@
...
+    timer:sleep(260),
+    Doc = kz_json:set_values([{<<"_id">>, Id}
+                             ,{<<"name">>, Name}
+                             ,{<<"status">>, <<"new">>}
+                             ,{<<"pvt_type">>, <<"replicator">>}
+                             ]
+                            ,kz_json:new()
+                            ),
+    {'ok', _} = kz_datamgr:save_doc(?KZ_REPLICATOR_DB, Doc),
+    io:format("created job ~s with name ~s\n",[Id, Name]).
```
jamesaimonetti: I would not use io:format in this module, return what's needed to the calling module and it can decide what to print to the SUP terminal

kazoo_replicator/src/replicator.erl
```diff
@@ -0,0 +1,183 @@
...
+    print_status(get_all_docs()).
+
+get_all_docs() ->
+    {'ok', Docs} = kz_datamgr:all_docs(?KZ_REPLICATOR_DB, ['include_docs']),
+    Docs.
+
+print_status([]) -> 'ok';
+print_status([Doc | Docs]) ->
+    Id = kz_doc:id(Doc),
+    JObj = kz_json:get_value(<<"doc">>, Doc),
```
jamesaimonetti: typed getters like `get_json_value` here

kazoo_replicator/src/replicator.erl
```diff
@@ -0,0 +1,183 @@
+%%%-----------------------------------------------------------------------------
+%%% @copyright (C) 2010-2025, 2600Hz
+%%% @doc
+%%% This Source Code Form is subject to the terms of the Mozilla Public
+%%% License, v. 2.0. If a copy of the MPL was not distributed with this
+%%% file, You can obtain one at https://mozilla.org/MPL/2.0/.
+%%%
+%%% @end
+%%%-----------------------------------------------------------------------------
+-module(replicator).
```
jamesaimonetti: Why not `kz_replicator`?

romanat-ooma: just didn't think of that, but love the change.

kazoo_replicator/src/replicator_util.erl
```diff
@@ -0,0 +1,181 @@
...
+                            ,{<<"source">>, url_section(get_from_section(Db, JObj))}
+                            ,{<<"target">>, url_section(get_to_section(Db, JObj))}
+                            ,{<<"continuous">>, kz_json:get_boolean_value(<<"continuous">>, JObj, 'true')}
+                            ,{<<"create_target">>, 'true'}
+                            ]),
+    save_doc(Url, Db, Opts, Doc).
+
+save_doc(Url, Db, Opts, Doc) ->
+    DocUrl = kz_term:to_binary([Url, <<"/">>, Db]),
+    _ = kz_http:put(DocUrl, [], Doc, Opts).
```
jamesaimonetti: should `kz_json:encode(Doc)` here, right? send JSON vs JObj?

romanat-ooma: oops didn't think of that.

kazoo_replicator/src/replicator.erl
romanat-ooma: changed to use lager instead of io:format

kazoo_replicator/src/replicator.erl
romanat-ooma: changed to kz_replicator :)

PR #1191, opened 2025-05-28
KZOO-634: A way to Monitor amqp channel consumption and total availability

romanat-ooma:
There is currently no way to monitor channel consumption and total availability. 
There needs to be a json output for sup command which lists no of channels available and total available channels.


kazoo_nodes/src/kz_node.erl
```diff
@@ -112,8 +112,11 @@ amqp_status_connection(Connection, JObj) ->
     Count = kz_amqp_assignments:channel_count(Connection),
     Broker = kz_amqp_connection:broker(Connection),
     BrokerZone = kz_amqp_connection:zone(Connection),
+    [{channel_max, ChannelMax}, _] = kz_amqp_connection:info(Connection),
 
     ConnStatus = kz_json:from_list([{<<"channel_count">>, Count}
+                                   ,{<<"channel_max">>, ChannelMax}
+                                   ,{<<"channel_remaining">>, ChannelMax - Count}
```
lazedo: remove channel_remaining, the info for getting that is already there

lazedo (changes requested):

kazoo_nodes/src/kz_node.erl
romanat-ooma: @lazedo removed :)

lazedo (approved):

kazoo_nodes/src/kz_node.erl
```diff
@@ -112,8 +112,10 @@ amqp_status_connection(Connection, JObj) ->
     Count = kz_amqp_assignments:channel_count(Connection),
     Broker = kz_amqp_connection:broker(Connection),
     BrokerZone = kz_amqp_connection:zone(Connection),
+    [{channel_max, ChannelMax}, _] = kz_amqp_connection:info(Connection),
```
lazedo:
```suggestion
    ChannelMax = props:get_value(channel_max, kz_amqp_connection:info(Connection, [channel_max])),
```

lazedo (changes requested):
lets be more specific on what we want to retrieve from kz_amqp_connection:info/2 and not rely on kz_amqp_connection:info/1 defaults that could interfere with the shape of the expected result

kazoo_nodes/src/kz_node.erl
romanat-ooma: Thank you @lazedo updated code as requested :)

lazedo (approved):

PR #1194, opened 2025-05-29
KZOO-636: secondary storage connections

jamesaimonetti:
Provide a mechanism to replicate saves/deletes to a secondary (or n-ary) connections.

Most operations are spawned to go to the remote connection, although db_create made sense to keep serial since subsequent doc changes would really benefit from the db existing on remote too, and race conditions could occur if creating the db and immediately trying to save a doc while the db_create on remote is still processing.


jamesaimonetti:
Schema updates here: https://github.com/2600hz/kazoo-crossbar/pull/717
Tests here: https://github.com/2600hz/kazoo-properly/pull/243
  - requires a separate Couch cluster be running apart from the `local` connection.

romanat-ooma (approved):

PR #1200, opened 2025-06-06
[5.5] KZOO-636: secondary storage connections

jamesaimonetti:
Backport of #1194

Provide a mechanism to replicate saves/deletes to a secondary (or n-ary) connections.

Most operations are spawned to go to the remote connection, although db_create made sense to keep serial since subsequent doc changes would really benefit from the db existing on remote too, and race conditions could occur if creating the db and immediately trying to save a doc while the db_create on remote is still processing.


jamesaimonetti:
CB: https://github.com/2600hz/kazoo-crossbar/pull/721

PR #1238, opened 2025-06-27
KZOO-626: update to erlazure API

jamesaimonetti:
No more gen_server to coordinate through, do uploads in parallel


jamesaimonetti:
Requires https://github.com/2600hz/kazoo5/pull/204

lazedo (approved):

PR #1252, opened 2025-07-18
KTEL-33: Set undefined if binding not found from map_results

dhuynh-Kazoo:
- binding map_results can return an empty list -> should convert it to `undefined`


kazoo_amqp/src/api/kapi_notifications.erl
```diff
@@ -2188,15 +2188,13 @@ api_definition(<<"cf_notification">>) ->
 api_definition(<<"number_feature_manual_action">>) ->
     number_feature_manual_action_definition();
 api_definition(TemplateId) ->
-    Definition = find_other_api_definition(TemplateId),
-    'true' = Definition =/= 'undefined',
```
icehess, line 2138: Revert this. The crash is here intentional. If you need `find_publisher` to not crash handle it that in that function instead in a try/catch.

icehess (changes requested):

kazoo_amqp/src/api/kapi_notifications.erl
```diff
@@ -3671,14 +3671,18 @@ publish_email_verification(API, ContentType) ->
...
+                      [PubFun
+                       || PubFun <- kazoo_bindings:map(binding_kapi_publisher(TemplateId), [TemplateId]),
+                          is_callable(PubFun)
+                      ];
+                  Api ->
+                      kapi_definition:publish_fun(Api)
+              end,
+    case PubFun1 of
+        [] -> 'undefined';
+        PubFun1 -> PubFun1
```
icehess:
```suggestion
    try api_definition(TemplateId) of
        [] ->
            case [PubFun
                  || PubFun <- kazoo_bindings:map(binding_kapi_publisher(TemplateId), [TemplateId]),
                     is_callable(PubFun)
                 ]
             of
                 [] -> 'undefined';
                 [Fun|_] -> Fun
             end;
        Api ->
            kapi_definition:publish_fun(Api)
    catch
        _T:_E:_ST -> 'undefined'
```

icehess (changes requested):

kazoo_amqp/src/api/kapi_notifications.erl
icehess: Also fix `cb_notifications:maybe_find_publisher/1`

PR #1253, opened 2025-07-18
KTIO-4: ignore numbers quantities for some carriers, add module_name

icehess:
This updates the service quantities calculation to not count numbers with some carrier module names in any classification.

Also adds module_name to number props when lookup account, so apps can use that when making decision. For example Jonny5 can ignore authz based on carrier module name.


icehess:
PRs:
- master:
    - https://github.com/2600hz/kazoo-core/pull/1253
    - https://github.com/2600hz/kazoo-jonny5/pull/32
    - https://gitlab.com/oomaforbin/oomacorp/2600hz/kazoo-trunkingio/-/merge_requests/15
    - https://github.com/2600hz/kazoo-properly/pull/248
- 5.5:
    - https://github.com/2600hz/kazoo-core/pull/1263
    - https://github.com/2600hz/kazoo-jonny5/pull/35
    - https://gitlab.com/oomaforbin/oomacorp/2600hz/kazoo-trunkingio/-/merge_requests/25
    - https://github.com/2600hz/kazoo-properly/pull/250
- 5.4:
    - https://github.com/2600hz/kazoo-core/pull/1261
    - https://github.com/2600hz/kazoo-jonny5/pull/33
    - https://gitlab.com/oomaforbin/oomacorp/2600hz/kazoo-trunkingio/-/tree/5.4
    - https://github.com/2600hz/kazoo-properly/pull/249
- meta-kazoo-5.4.18:
    - https://github.com/2600hz/kazoo-core/pull/1262
    - https://github.com/2600hz/kazoo-jonny5/pull/34
    - https://gitlab.com/oomaforbin/oomacorp/2600hz/kazoo-trunkingio/-/merge_requests/24
    - no properly tag exists for this release

icehess:
~WIP btw, please do not merge yet~ -- ready to merge.

kazoo_services/src/kz_services_quantities.erl
```diff
@@ -559,11 +559,16 @@ calculate_phone_numbers_updates(JObj, Updates) ->
...
-    case kz_json:get_value([<<"pvt_features">>, <<"local">>], JObj) =:= 'undefined' of
+    case phone_number_is_unbillable(JObj) of
         'false' -> Updates;
         'true' -> [{Key, 1} | Updates]
     end.
 
+-spec phone_number_is_unbillable(kz_json:object()) -> boolean().
+phone_number_is_unbillable(JObj) ->
+    not kz_json:is_defined([kzd_phone_numbers:pvt_features_path(), <<"local">>], JObj)
+        andalso <<"knm_trunkingio">> =/= kz_json:get_ne_binary_value(<<"pvt_module_name">>, JObj).
```
jamesaimonetti: should this pull from a system_config doc instead? maybe a `lists:member(pvt_module_name, kapps_Config:get(numbers, unbillable, [])` with ops instructions to include knm_tio?

icehess: yes, eventually, if everyone agree about this change in general.

lazedo: why not use use knm_carriers:is_number_billable/1 ?

icehess: because i didn't even noticed that function existed :)) this actually make it faster, even in jonny5.

SO i don't know what is better here. doing this completely turns off billing for all accounts. it is not even possible for reseller to charge.

I don't know if we ever want to overrides charges for reseller (which can be done in kazoo-trunkingio), but allow reseller to bill its subaccounts as they desire.

jamesaimonetti: if a call is using a TIO number, and TIO is handling billing, no reason for kazoo to track billing too, right? so i'm in agreement to use the is_number_billable/1 and let knm_trunkingio decide (you have the whole PN record so access to account info if you want to toggle reseller vs account billing, for instance).

lazedo:
why are trying to add all these changes to is_number_billable ?

kazoo_services/src/modules/kz_services_im.erl
```diff
@@ -51,11 +51,42 @@ fetch(?NE_BINARY=AccountId) ->
...
+            reseller_fetch(Services);
+        Plan ->
+            Plan
+    end;
+reseller_fetch(Services) ->
+    {AccountIM, AccountCaches} = merge_plans(Services),
+    cache(kz_services:account_id(Services), AccountCaches, AccountIM).
+
+-spec merge_plans(kz_services:services()) -> {kz_json:object(), list()}.
+merge_plans(Services) ->
```
lazedo: why are you making these changes ?

kazoo_services/src/modules/kz_services_im.erl
```diff
@@ -51,11 +51,42 @@ fetch(?NE_BINARY=AccountId) ->
             Plan
     end;
 fetch(Services) ->
+    {AccountIM, AccountCaches} = merge_plans(Services),
+    MergeFromReseller = kapps_config:is_true(?IM_CONFIG_CAT, <<"merge_plan_from_reseller">>, 'false'),
+    case kz_services_reseller:is_reseller(Services)
+        andalso MergeFromReseller
+    of
+        'true' ->
+            cache(kz_services:account_id(Services), AccountCaches, AccountIM);
```
lazedo: this logic looks inverted ?

kazoo_numbers/src/carriers/knm_carriers.erl
```diff
@@ -365,14 +365,21 @@ direct_reseller_id(Options) ->
 %% @doc Returns whether carrier handles numbers local to the system.
 %% @end
 %%------------------------------------------------------------------------------
--spec is_number_billable(knm_phone_number:pn_record()) -> boolean().
+-spec is_number_billable(knm_phone_number:pn_record() | kz_term:ne_binary()) -> boolean().
```
lazedo: wouldn't it be easier to handle the binary case here and fetch the number record and apply it instead of all the changes that you're trying to introduce?

kazoo_services/src/kz_services_quantities.erl
icehess: kk

kazoo_numbers/src/carriers/knm_carriers.erl
icehess: was trying to avoid a bit more expensive knm_phone_number:fetch, since this function is also called by j5_authz_req.

kazoo_services/src/modules/kz_services_im.erl
icehess: reverted

kazoo_services/src/modules/kz_services_im.erl
icehess: reverted

kazoo_services/src/kz_services_quantities.erl
```diff
@@ -559,11 +559,18 @@ calculate_phone_numbers_updates(JObj, Updates) ->
...
         'true' -> [{Key, 1} | Updates]
     end.
 
+-spec phone_number_is_billable(kz_json:object()) -> boolean().
+phone_number_is_billable(JObj) ->
+    case kz_json:is_defined([kzd_phone_numbers:pvt_features_path(), <<"local">>], JObj) of
+        'true' -> 'false';
+        'false' -> knm_numbers:is_number_billable(JObj)
+    end.
+
```
lazedo: doesn't knm_local return false ?

icehess: done. was not sure which way was better. changed it to only `knm_numbers:is_number_billable/1` now.

jamesaimonetti (approved):

